PR #94, opened 2014-12-11
Make Movie script handles units for scalebar

will-moore:
This fixes the Make Movie script to work with units changes.

All other scripts seem to have been fixed already.

To test, check that you can create a movie with a scalebar.


jburel:
It will probably be necessary to adjust the scale bar to the units used. 
It is still in microns.
This can be done in a follow-up PR
Also noticed on make_movie.figure that ```colour``` is used instead of  ```color```

jburel:
@gusferguson: Ok to merge?

gusferguson:
@jburel 
Fine by me if the little things are noted to be done at another stage.

jburel:
Other points to be addressed in another PR. Merging

PR #95, opened 2015-01-05
Movies saved with specified filename

will-moore:
Fixes naming of Movies created with Make_Movie script, as reported in https://github.com/openmicroscopy/openmicroscopy/pull/3282#issuecomment-67302122

Test in Insight and Web:
 - Movie will now be saved (as attachment to image) according to the file name specified in the script dialog.
 - NB: Insight and web still have different policies for the default name in the Make Movie dialog when handling images named with path/to/image.tiff

PR #96, opened 2015-01-23
Movies saved with specified filename (rebased onto dev_5_0)

will-moore:
This is the same as gh-95 but rebased onto dev_5_0.

----

Fixes naming of Movies created with Make_Movie script, as reported in https://github.com/openmicroscopy/openmicroscopy/pull/3282#issuecomment-67302122

Test in Insight:
 - Movie will now be saved (as attachment to image) according to the file name specified in the script dialog.
 - NB: Insight and web still have different policies for the default name in the Make Movie dialog when handling images named with path/to/image.tiff


will-moore:
--rebased-from #95

dominikl:
Works like expected. Tested with Insight only, 'Make Movie' script not available on 5.0 Web, but that's expected according to @will-moore .

PR #97, opened 2015-02-09
Move import at the top of the file

sbesson:
This should allow future PR to pass Travis

PR #98, opened 2015-02-11
Move import at the top of the file (rebased onto dev_5_0)

sbesson:
This is the same as gh-97 but rebased onto dev_5_0.

----

This should allow future PR to pass Travis


sbesson:
--rebased-from #97

mtbc:
Good to merge.

PR #100, opened 2015-03-20
image._re.close() during Batch_Image_Export

will-moore:
From discussion at https://github.com/openmicroscopy/openmicroscopy/pull/3639

We now close the rendering engine for each image after all the planes for that image have been saved.

To test, simply confirm that the Batch Image Export script is still working.


mtbc:
Good to merge.

PR #103, opened 2015-07-01
Script for calculating min/max

joshmoore:
After importing with `--skip=all` the rendering settings for an image can be quite unattractive. If the range is large, finding a good setting manually can be problematic. This calculates the min/max per `Channel` and optionally creates or updates the `StatsInfo` for each `Channel`.

cc: @manics @jburel @will-moore @sbesson

### DIFF
--- a/omero/util_scripts/Min_Max.py
+++ b/omero/util_scripts/Min_Max.py
@@ -22,6 +22,9 @@
 ------------------------------------------------------------------------------
 """
 
+import omero.all
+
+from omero import MissingPyramidException
 from omero.gateway import BlitzGateway
 from omero.model import StatsInfoI
 from omero.rtypes import rdouble, rlong, rstring
@@ -55,9 +58,14 @@ def calcStatsInfo(conn, imageId):
     sizeY = oldImage.getSizeY()
 
     # check we're not dealing with Big image.
-    rps = oldImage.getPrimaryPixels()._prepareRawPixelsStore()
-    bigImage = rps.requiresPixelsPyramid()
-    rps.close()
+    bigImage = False
+    try:
+        rps = oldImage.getPrimaryPixels()._prepareRawPixelsStore()
+        bigImage = rps.requiresPixelsPyramid()
+        rps.close()
+    except MissingPyramidException:
+        bigImage = True
+
     if bigImage:
         raise Exception((
             "This script does not support 'BIG' images such as Image ID: "
@@ -108,7 +116,7 @@ def processImages(conn, scriptParams):
     images, logMessage = script_utils.getObjects(conn, scriptParams)
     message += logMessage
     if not images:
-        return None, None, message
+        raise Exception("No images found")
     imageIds = sorted(set([i.getId() for i in images]))
 
     globalmin = defaultdict(list)
--- a/omero/util_scripts/Min_Max.py
+++ b/omero/util_scripts/Min_Max.py
@@ -25,6 +25,7 @@
 from omero.gateway import BlitzGateway
 from omero.model import StatsInfoI
 from omero.rtypes import rdouble, rlong, rstring
+from omero.util.text import TableBuilder
 from omero.util.tiles import TileLoop, TileLoopIteration
 
 import omero.scripts as scripts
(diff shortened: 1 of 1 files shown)


jburel:
For performance, It might be actually better to calculate the rendering settings for the default plane (z =SizeZ/2, t=0)
WindowStart will be set to the min of that plane (one per channel) and WindowEnd = max of that plane 
this means that we will only scan few planes.

joshmoore:
Is it useful to allow either or?

jburel:
Both will be useful.
The second one follows more the aim of  ```--skip=all``` in my opinion.
i.e. we still offer a good visualisation experience at a low cost.

joshmoore:
Thoughts on the argument name and the default "off" or "on"?

manics:
Another option to speed things up: randomly pick `N` tiles across all C/Z/T/X/Y for all wells, and use a single set of rendering settings for the whole plate.

jburel:
The "shared" rendering settings will require some adjustments in the way we use them to generate thumbnails or display images.

joshmoore:
> The "shared" rendering settings will require some adjustments in the way we use them to generate thumbnails or display images.

Hadn't thought about having anything shared. I'll probably leave that for further efforts.

joshmoore:
https://github.com/joshmoore/scripts/commit/4927d84a7c3ccf6c71ca3578768210c59b4a86d0 is a start on "default" and "random".

jburel:
Going back to the script, I am now wondering if we should not "run" it when calculating the min/max during the rendering steps

joshmoore:
Note: @emilroz had also started working on a version for big images while this is for HCS images. I haven't found time to unify yet.

jburel:
Could such option be part of ``omero-cli-render``?

joshmoore:
Certainly, that would match the strategy we've used with populate metadata as well:

```
 - library
    |_ cli
    \_ script
```

cc: @sbesson

omero/util_scripts/Min_Max.py
```diff
@@ -0,0 +1,259 @@
...
+    if debug:
+        print "Loaded %s tile(s) (%s bytes)" % (tile_count, byte_count)
+    return statsInfos
+
+
+def processImages(conn, scriptParams):
+    """
+    Process the script params to make a list of channel_offsets, then iterate
+    through the images creating a new image from each with the specified
+    channel offsets
```
mtbc, line 128: May be out of date.

omero/util_scripts/Min_Max.py
```diff
@@ -0,0 +1,259 @@
...
+            shuffle(copy)
+            zctMap[c] = copy
+
+    def channelGen():
+        byte_count = 0
+        tile_count = 0
+        pixels = oldImage.getPrimaryPixels()
+        rv = dict()
+        dt = pixels.getTile(0, 0, 0, (0, 0, 16, 16)).dtype
+        tile_min = iinfo(dt).max  # Everything is less
```
joshmoore, line 103: Note: this fails for floating point data types.

omero/util_scripts/Min_Max.py
```diff
@@ -0,0 +1,259 @@
...
+                tile_count += 1
+            rv[c] = (tile_min, tile_max)
+        yield rv, byte_count, tile_count
+
+    statsInfos = dict()
+    for x, byte_count, tile_count in channelGen():
+        statsInfos.update(x)
+
+    if debug:
+        print "Loaded %s tile(s) (%s bytes)" % (tile_count, byte_count)
```
pwalczysko, line 120: This one and other similar print statements error in my hands. Is it because of the python3 environment I am using ?

will-moore: Yes, just use `print()` like 
```
print("Loaded %s tile(s) (%s bytes)" % (tile_count, byte_count))
```

jburel:
--exclude

joshmoore:
This didn't make it into the first hour of OMERO PR reviews, but I'm closing nonetheless. I do want my award, though, for the oldest outstanding PR. :wink:

chris-allan:
> This didn't make it into the first hour of OMERO PR reviews, but I'm closing nonetheless. I do want my award, though, for the oldest outstanding PR. 😉

👑 Moore I of Mount PR.

PR #106, opened 2015-11-26
Keep Dataset as default

dominikl:
See [Trello - scripts: review scripts with RE](https://trello.com/c/H3lObNDY/10-scripts-review-scripts-with-re)

Test: Open Dataset-To-Plate script dialog; make sure the "Remove images from Dataset" checkbox is _not_ ticked by default.

### DIFF
--- a/omero/util_scripts/Dataset_To_Plate.py
+++ b/omero/util_scripts/Dataset_To_Plate.py
@@ -327,7 +327,7 @@ See http://help.openmicroscopy.org/utility-scripts.html""",
             " screen or ID of existing screen"""),
 
         scripts.Bool(
-            "Remove_From_Dataset", grouping="7", default=True,
+            "Remove_From_Dataset", grouping="7", default=False,
             description="Remove Images from Dataset as they are added to"
             " Plate"),
 


will-moore:
What's the logic behind this change? It just seems to me that in general it's not a good idea to have images linked to Wells AND Datasets?

dominikl:
As far as I remember the problem was that this is an action which cannot be reverted, so removing the images as default option is a bit dangerous.

dominikl:
Hold off til next client meeting? /cc @gusferguson

gusferguson:
@dominikl @will-moore 

Tested using OMERO.insight-5.2.0-216-c50d792-ice35-b168-mac user-3

Behaves as expected.
I still think this is the correct behaviour as it is consistent with good UI practice and is safest for the majority of users.
Good to merge.

will-moore:
@gusferguson I don't think it's better to keep images linked to Dataset AND to Well, since users will not understand this and none of our code expects this. E.g. try to move the new Plate to a different group:

screen shot 2015-11-27 at 10 58 42

Or try to delete one of the images remaining in the Dataset:

screen shot 2015-11-27 at 11 04 30

gusferguson:
@will-moore 

I can see it causes problems - but it is already possible - just not as default.

It just goes against all principles to have an irreversible "destructive" action as a default - with no warning - particularly as you point out - when users are extremely unlikely to understand what actually happens in reality when they run the script.

If you feel strongly about this then just leave unselected as default, and we can add a warning dialogue in a different PR with humble apologies to @dominikl  for wasting his time.

dominikl:
Well, a change from a True to a False wasn't such a big waste of time ;-) But I'd also suggest to add some kind of warning in this case.

PR #107, opened 2016-02-04
Populate metadata

will-moore:
As discussed https://github.com/openmicroscopy/openmicroscopy/pull/4455#issuecomment-179714477 Now that we support visualisation of bulk annotations in web (and Insight via tooltip) we want to make it easier for users to attach bulk annotations.

This PR copies the Glencoe script from https://github.com/glencoesoftware/omero-user-scripts/blob/master/util_scripts/Populate_Metadata.py and updates it a bit:
- Add (hopefully) useful description
- Use File_Annotation ID for picking files instead of File ID (web and Insight UI allow for picking File Ann)
- Make File_Annotation optional - If not set, we simply pick a .csv file from the Plate

To test, use Plate and csv files described https://github.com/openmicroscopy/ome-internal/blob/master/testing_scenarios/BulkAnnotations.txt
- Import Plate, attach test_plate.csv to it
- Select Plate and run script > Import Scripts > Populate_Metadata
- Read description. Makes sense?
- Run (don't need to choose File Annotation - will pick the csv automatically).
- Also, if multiple csv attached, you can pick the one you want from right panel and this will be used to populate the "File Annotation" parameter.
- When the script completes, selecting Wells should show the Bulk Annotation in right panel (as in https://github.com/openmicroscopy/openmicroscopy/pull/4446).

screen shot 2016-02-04 at 15 27 54

screen shot 2016-02-04 at 15 26 12

### DIFF
--- a/util_scripts/Populate_Metadata.py
+++ b/util_scripts/Populate_Metadata.py
@@ -34,14 +34,14 @@ from omero.util.populate_roi import DownloadingOriginalFileProvider
 from omero.util.populate_metadata import ParsingContext
 
 
-def get_original_file(conn, object_type, plate_id, file_id):
+def get_original_file(conn, object_type, object_id, file_id):
     if object_type == "Plate":
-        omero_object = conn.getObject("Plate", int(plate_id))
+        omero_object = conn.getObject("Plate", int(object_id))
         if omero_object is None:
             sys.stderr.write("Error: Plate does not exist.\n")
             sys.exit(1)
     else:
-        omero_object = conn.getObject("Screen", int(plate_id))
+        omero_object = conn.getObject("Screen", int(object_id))
         if omero_object is None:
             sys.stderr.write("Error: Screen does not exist.\n")
             sys.exit(1)
--- a/util_scripts/Populate_Metadata.py
+++ b/util_scripts/Populate_Metadata.py
@@ -26,7 +26,7 @@ import omero
 from omero.gateway import BlitzGateway
 from omero.rtypes import rstring
 import omero.scripts as scripts
-from omero.model import PlateI
+from omero.model import PlateI, ScreenI
 
 import sys
 
@@ -34,8 +34,8 @@ from omero.util.populate_roi import DownloadingOriginalFileProvider
 from omero.util.populate_metadata import ParsingContext
 
 
-def get_original_file(conn, plateId, fileId):
-    plate = conn.getObject("Plate", int(plateId))
+def get_original_file(conn, plate_id, file_id):
+    plate = conn.getObject("Plate", int(plate_id))
     if plate is None:
         sys.stderr.write("Error: Object does not exist.\n")
         sys.exit(1)
@@ -44,7 +44,7 @@ def get_original_file(conn, plateId, fileId):
         if isinstance(ann, omero.gateway.FileAnnotationWrapper):
             print "File ID:", ann.getFile().getId(), ann.getFile().getName(),\
                 "Size:", ann.getFile().getSize()
-            if (ann.getFile().getId() == int(fileId)):
+            if (ann.getFile().getId() == int(file_id)):
                 file = ann.getFile()._obj
     if file is None:
         sys.stderr.write("Error: File does not exist.\n")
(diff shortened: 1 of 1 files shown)


will-moore:
cc @emilroz I've copied your script here and made a few changes (see above).

pwalczysko:
The output on the right-hand side is a bit repetitive. Not sure why. I have admittedly ran the script twice (deleting the `Bulk_annotation` attachment in between of the two script executions) but still, some lines seem to be there 4 times and more.
screen shot 2016-02-05 at 12 01 22

pwalczysko:
Re: https://github.com/ome/scripts/pull/107#issuecomment-180320777 Clarified with @will-moore that the duplication came from the fact that there was another ``Bulk_annotation` attachment on the containing Screen left from previous testing. When this attachment was deleted, the duplication vanished. 

This poses an important question of `what to do when there is multiple exemplaries of the attachment` -> tactics might differ (even between Web and Insight), but definitely the display should be separate for each attachment in the right-hand pane display, similarly to the Key-Value pairs.

pwalczysko:
screen shot 2016-02-05 at 12 31 47

I think a hint about how do I construct the csv in case it is attached to a screen would be very useful in the description of the script (see above).
This `screen` workflow works for me having following .csv (see screenshot below) where the `Plate` column actually contains the name of the plate in OMERO (it is just misformatted by Excel as a number, see second screenshot below)

screen shot 2016-02-05 at 12 36 58
screen shot 2016-02-05 at 12 38 58

emilroz:
@will-moore the script looks good to me.
@pwalczysko I think that an example `.csv` file would be better then a very long description. As you've noticed the formatting plays also very important role here, as we are really only prepared now to take properly formatted comma separated value files anything else will fail silently - you will get a single column table as an output and as a result no `bulk annotations` on the images.

pwalczysko:
@emilroz : Agree, an example file is better. How shouild the user find such a file ?

joshmoore:
Will a link to a file be clickable in web & insight? If so, I'd think a link to a [sample image](http://downloads.openmicroscopy.org/images) (perhap a fake?) which includes a CSV with a readme on how to install it would be useful?

cc: @sbesson @simleo @manics -- This would also be a first candidate for https://trello.com/c/9aEYto1f/217-table-import-prototype

will-moore:
@joshmoore Apologies for the delay. Unfortunately web scripts dialog doesn't make links clickable at the moment. However, I think a non-clickable link to proper docs and example csv is still the best option.
We just need to get docs in place so we have the url. Don't know if we can do this in time for this PR: @gusferguson?

gusferguson:
@will-moore @joshmoore - what were you thinking in terms of User Help documentation? Easy enough to have an example doc in the resources part of the User Help - that will give an immutable URL for downloads and whatever/wherever there is something in the actual User Help.

joshmoore:
I'll leave you guys to discuss what's doable for 5.2.2. We can capture any extras in a card for later.

NB: @will-moore, sorry for having missed it earlier, could you possibly rebase this onto of https://github.com/joshmoore/scripts/commit/47cd2ff248e3ed1649f298fa03258d33265ccd9f (i.e. replacing your 7298da80977fe7927c28accf014767938e6280e9 with commits from @emilroz so that we have the history. (I've included https://github.com/joshmoore/scripts/commit/a0dd39261f54599d9e795d6137d5cfb89c9ed599 on top to do the rename unless you want to do that yourself)

gusferguson:
@will-moore @joshmoore 
Confirmed that I will be putting the Help for this script in the Using Scripts page with a #metadata anchor, so the URL: 
http://help.openmicroscopy.org/scripts.html#metadata
will be the one to use.

jburel:
The script does not follow the approach used for ids in other scripts i.e. ids as `rlist` not `rstring`
This was the source of the error in insight https://github.com/openmicroscopy/openmicroscopy/pull/4487
and it also means that the parsing will now happen in the script itself if multiple ids are passed

cc @will-moore @emilroz

gusferguson:
Has this been tested with Insight? 
I am getting errors trying to run it on eel using today's build. 
Works fine in Web.

gusferguson:
Actually not running in web now.

jburel:
@emilroz: are you okay if we change the type for IDs i.e. list instead of String.
This will require some re-tests of the scripts
The only limitation is that the "FileAnnotation" option will potentially not work when multiple objects are selected (unless it is linked to all the objects)

jburel:
@gusferguson: insight PR is https://github.com/openmicroscopy/openmicroscopy/pull/4487. I pushed a fix this am (not included in the build)

gusferguson:
@jburel +1

will-moore:
Updated the URL in the description as discussed.
Also switched to using List for IDs. We only support a single ID. Warning is printed to output if multiple IDs are given - we just use the first.
Added a response message to display when the script completes.

To test, select multiple Screens or Plates. Should see warning in Info when complete.
Also if completes, should see a simple message:

screen shot 2016-02-22 at 23 26 02

jburel:
Message displayed if multiple ids selected:
for plate

```
{'IDs': [226L, 8L], 'Data_Type': 'Plate'}
WARNING: Multiple IDs not currently supported
    Only using the first ID: 226
Listing files on Plate 226...
```

for screen

```
{'IDs': [801L, 1151L], 'Data_Type': 'Screen'}
WARNING: Multiple IDs not currently supported
    Only using the first ID: 801
Listing files on Screen 801...
```

jburel:
Log in as user-4, screen `Bulk_Annotation_test`
`WARNING:omero.util.populate_metadata:PlateColumn is unimplemented`

jburel:
Warning introduced in https://github.com/joshmoore/openmicroscopy/commit/00d89c64b22bd8e6adaed09f2d806861c3a5bb25
cc @joshmoore 
Web indicated that the result has been created. No message in insight only the error button.
This will have to be improved post release in insight

jburel:
The script works as expected now. I will retest of insight w/o https://github.com/openmicroscopy/openmicroscopy/pull/4487  included (changes no longer required)

emilroz:
@jburel I'm fine with changing the parameter type.

Edit: added `fine` :)

jburel:
@emilroz: Thanks, that's the best option. I have closed my insight PR.

will-moore:
@joshmoore That's the rebasing done that you asked for in https://github.com/ome/scripts/pull/107#issuecomment-187079776, including the renaming commit. I checked that the script is unchanged from before, (we just have the commit history) so it doesn't need retesting.

jburel:
Tested w/o insight changes included, it works fine.

joshmoore:
Note: without the `git mv` this is now the only script we ship outside of `omero`.

will-moore:
Sorry @joshmoore I had cherry-picked your move commit locally but hadn't pushed it. Done now.
Ooops - I see it's just been merged already.

will-moore:
I'll open a new PR with the move commit.

```
* 1155250 (HEAD -> Populate_Metadata, gh/Populate_Metadata) Rename from util_scripts
```

jburel:
@will-moore: A PR has now been opened.

will-moore:
OK - e-mail delay is annoying! Closed https://github.com/ome/scripts/pull/110

joshmoore:
Sorry, @will-moore. I should have commented here (I had even thought about it!)

PR #108, opened 2016-02-25
Place Populate_Metadata.py under import_scripts

joshmoore:
Follow-on to gh-107 with a renaming based on the standard
ome/scripts scheme.


jburel:
Thanks @joshmoore.
@gusferguson this will match the screenshots you have prepared so nothing to adjust.

jburel:
Merging. We will double-check with rc.

PR #114, opened 2016-05-17
batch export big images

will-moore:
This improves handling of Big images in Batch Image Export script.

Previously, the first check was the 'omero.client.download_as.max_size' setting (regardless of which format you were exporting). And if smaller than this limit then OME-TIFF export would go ahead, even if image was a 'Big' (tiled) image that is not supported by OME-TIFF exporter service.

Now, we first check the export format.
If OME-TIFF, then we check if it's a 'Big' (tiled) image.
Otherwise we check if it's under the 'omero.client.download_as.max_size' limit (12k \* 12k by default).

To test:
- Try to export a single image (over 12k \* 12k) as JPG / PNG. Should get sensible message in scripts activities and log.
- Try to export a single 'Big' image at OME-TIFF (can use same image as above). Should get a different message in scripts activities and log.
- Try to export multiple images as OME-TIFF or PNG/JPG. If some are too big these will give message in log. If none are export (all too big) should see reasonable message. Otherwise will get zip.

screen shot 2016-05-17 at 16 47 40

### DIFF
--- a/omero/export_scripts/Batch_Image_Export.py
+++ b/omero/export_scripts/Batch_Image_Export.py
@@ -392,25 +392,33 @@ def batchImageExport(conn, scriptParams):
     # do the saving to disk
 
     for img in images:
+        log("Processing image: ID %s: %s" % (img.id, img.getName()))
         pixels = img.getPrimaryPixels()
         if (pixels.getId() in ids):
             continue
         ids.append(pixels.getId())
-        sizeX = pixels.getSizeX()
-        sizeY = pixels.getSizeY()
-        if sizeX*sizeY > size:
-            log("  ** Can't export a 'Big' image to %s. **" % format)
-            if len(images) == 1:
-                return None, "Can't export a 'Big' image to %s." % format
-            continue
-        else:
-            log("Exporting image as %s: %s" % (format, img.getName()))
 
         if format == 'OME-TIFF':
-            saveAsOmeTiff(conn, img, folder_name)
-        else:
             if img._prepareRE().requiresPixelsPyramid():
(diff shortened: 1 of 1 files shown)


gusferguson:
@will-moore 

Tested with cowfish user-3 read-only-1

Behaves as expected.
Good to merge.

jburel:
What happens if I select a large number of them?
Did you try that @gusferguson ?

will-moore:
This PR isn't trying to improve handling of large number of images (which is why I didn't ask for that in the tests). This is trying to fix the OME-TIFF export going ahead with a 'Big' image & getting an error (not supported).

gusferguson:
I tried with about 5 I think.

jburel:
My point about large number is that `big images+few images` might become an issue.
Opening to "big image" might lead to other issues that need to be looked at.

will-moore:
The point of this PR was simply to fix the logic of handling 'Big' images that are less than 12k \* 12k. It doesn't attempt to address any issues with downloads of large amounts of data.
If you try to zip up a large amount of data, it will work up to some limit which we haven't tried to determine and which will be different on different hardwear etc. While it may be interesting to know that, it doesn't really affect the logic in this PR.
If it's something we want to look into, lets make a trello card somewhere. However, in general batch_image_export script isn't affected by large data as much as Download from web (timeout issues etc).

jburel:
I understand, I am just raising a point since we have not been too careful in the past with some scripts e.g. projection
If you can add a card, that will be great

will-moore:
Card at https://trello.com/c/JvLHywJo/127-batch-image-export-of-large-amount-of-data

jburel:
Thanks

PR #116, opened 2016-06-16
Batch Image Export resize

will-moore:
This fixes a tiny bug when exporting images with Zoom > 100%

To test:
- Run Batch Image Export script on an image and choose a "Zoom" parameter of E.g. 200%.
- Images should be exported as jpeg/png but resized by the chosen amount (previously script failed).


gusferguson:
@will-moore 

Behaves as expected for default and zoom values.
Good to merge

PR #118, opened 2016-08-09
Drop Python 2.6 from the Travis matrix

sbesson:
With flake8 3.0.0, Python 2.6 support has been dropped. Instead of capping flake8, this commit cuts 2.6 from the supported matrix and should restore the build.


manics:
Looks good to me

PR #119, opened 2016-10-04
Combine images pixel size

will-moore:
See https://trello.com/c/e8t6LrDB/197-combine-images-script-ignores-pixel-size

If images used in the CombineImages script have pixel sizes set, the script will pick these sizes and set them on the new image.

We need to test combining images that have:
- No pixel sizes set
- Same pixel sizes set in both/all
- Different pixel sizes in both images

We can set pixel sizes on test images using

```
$ omero shell --login
conn = omero.gateway.BlitzGateway(client_obj=client)
from omero.model.enums import UnitsLength
i = conn.getObject("Image", 1234)
u = omero.model.LengthI(9.8, UnitsLength.ANGSTROM)
p = i.getPrimaryPixels()._obj
p.setPhysicalSizeX(u)
p.setPhysicalSizeY(u)
conn.getUpdateService().saveObject(p)
```

To test:
- Pick a couple of single-plane images of same width & height. Easiest if these have no pixel sizes set initially.
- Try combining these. New image will have no pixel sizes set
- Set pixel size X on both images to the same value
- Set pixel size Y on just one image
- Try combining images. Size X and Size Y should be set on new image.
- Finally set Size Y on the other image, to a different value than above.
- Try combining images. Size Y should not be set on new image.
- Check info for each script run above - should see basic info like:

```
Pixel size mismatch! Won't be set:  1.5 MICROMETER 1.25 MICROMETER
Setting pixel sizes... X: None Y: 1.5 MICROMETER
```

### DIFF
--- a/omero/util_scripts/Combine_Images.py
+++ b/omero/util_scripts/Combine_Images.py
@@ -309,6 +309,7 @@ def pickPixelSizes(pixelSizes):
             # compare - if different, return None
             if (pixSize.getValue() != px.getValue() or
                     pixSize.getUnit() != px.getUnit()):
+                print "Pixel size mismatch! Won't be set: ", pixSize, px
                 return None
     return pixSize
 
@@ -441,7 +442,7 @@ def makeSingleImage(services, parameterMap, imageIds, dataset, colourMap):
     # Set pixel sizes if known
     pixSizeX = pickPixelSizes(pixelSizes['x'])
     pixSizeY = pickPixelSizes(pixelSizes['y'])
-    print pixSizeX, pixSizeY
+    print "Setting pixel sizes... X: %s Y: %s" % (pixSizeX, pixSizeY)
     if pixSizeX is not None or pixSizeY is not None:
         # reload to avoid OptimisticLockException
         pixels = services["queryService"].get('Pixels', pixels.id.val)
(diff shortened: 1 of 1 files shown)


pwalczysko:
On unset images, the output of info is  `Setting pixel sizes... X: None Y: None`
On setting the two X pixelsizes and one Y pixelsize the info is ``Setting pixel sizes... X: 9.8 ANGSTROM Y: 9.8 ANGSTROM`

On setting the two X pixelsizes and two numerically different Y pixelsizes (the same unit), info is `Pixel size mismatch! Won't be set:  15.0 ANGSTROM 9.8 ANGSTROM
Setting pixel sizes... X: 9.8 ANGSTROM Y: None`

Maybe reformulate the last output to `....mismatch! The mismatched dimension(s) will not be set:....` (get rid of the `won't` and be more precise)

Will do one more test mismatching the units (not the values).

pwalczysko:
Mismatching units (same values) do not get the mismatched dimension set as expected, with analogous info warning.

pwalczysko:
In summary, all good here except one RFE, see paragraph 3 of [comment](https://github.com/ome/scripts/pull/119#issuecomment-251645087)

pwalczysko:
Tweek of the warning works as expected.

Ready to merge.

PR #129, opened 2017-02-08
IDR: Remove all scripts

joshmoore:
This brings the `metadata` branch first up to date with the current state of the scripts repo for 5.2 and then removes all scripts so that they don't have to be disabled for every IDR deployment.


joshmoore:
Travis failure is Py26 flake issue:

```
    self._parameters = utils.parameters_for(self)
  File "/home/travis/virtualenv/python2.6.9/lib/python2.6/site-packages/flake8/utils.py", line 297, in parameters_for
    parameters = collections.OrderedDict([
AttributeError: 'module' object has no attribute 'OrderedDict'
```

joshmoore:
I'm going to merge this for idr 0.3.4. If it turns out this is too confusing or problematic, I'll likely force-push it away.

joshmoore:
Note: I will set metadata54 to point at this commit as well.

cc: @jburel

jburel:
Long term we should look into a client option "disable script" so it is easier to configure

joshmoore:
Minimally as part of the public-user system, definitely.

PR #134, opened 2017-03-03
Move_Annotations

will-moore:
Script to Select Screens, Plates, or Wells and move Annotations from Images onto their parent Wells.
See https://trello.com/c/gasafKWk/291-update-script-wellsample-well-annotation-migration

To test:

- Add a bunch of Annotations to Images in Wells. Pick some Wells that have multiple fields.
     - Add tags, ratings, map annotations, comments etc. 
- Select Screens, Plates or Wells and run script (check that selected S/P/W are populated in dialog)
- Choose to filter annotations by Type or Namespace. (can test this using ```openmicroscopy.org/omero/client/mapAnnotation``` for Map Annotations).
- Test "Remove Annotations" removes annotations from Images if checked.

screen shot 2017-03-03 at 14 59 17

### DIFF
--- a/omero/util_scripts/Move_Annotations.py
+++ b/omero/util_scripts/Move_Annotations.py
@@ -25,6 +25,7 @@ from omero.model import ExperimenterI, \
                         ImageAnnotationLinkI, \
                         WellAnnotationLinkI, \
                         WellI
+from omero.sys import ParametersI, Filter
 from omero.rtypes import rstring, rlong
 from omero.constants.metadata import NSINSIGHTRATING
 
@@ -38,20 +39,27 @@ ANN_TYPES = {
 }
 
 
-def log(*args):
+def log(text):
     """Handle logging statements in a single place."""
-    print args
+    print text
 
 
 def move_well_annotations(conn, well, ann_type, remove_anns, ns):
     """Move annotations from Images in this Well onto the Well itself."""
-    log("Processing Well:", well.id, well.getWellPos())
+    log("Processing Well: %s %s" % (well.id, well.getWellPos()))
     iids = [wellSample.getImage().id for wellSample in well.listChildren()]
-    log("  Image IDs:", iids)
+    log("  Image IDs: %s" % iids)
     if len(iids) == 0:
         return 0
 
-    old_links = list(conn.getAnnotationLinks('Image', iids, ns=ns))
+    # Params to query links. If not Admin, only work with our own links
+    params = ParametersI()
+    if not conn.isAdmin():
+        params.theFilter = Filter()
+        params.theFilter.ownerId = rlong(conn.getUserId())
+
+    old_links = list(conn.getAnnotationLinks('Image', iids,
+                                             ns=ns, params=params))
 
     # Filter by type
     old_links = [l for l in old_links
@@ -60,27 +68,34 @@ def move_well_annotations(conn, well, ann_type, remove_anns, ns):
 
     link_ids = [l.id for l in old_links]
 
+    def get_key(ann_link, with_owner=False):
+        # We use ann's 'key' to avoid adding duplicate annotations
+        # Key includes link owner (allows multiple links with different owners)
+        ann = ann_link.child
+        return "%s_%s" % (ann_link.details.owner.id.val, ann.id.val)
+
     links_dict = {}
-    # If current user is Admin, we preserve ownership of annotation links
-    # Therefore, annotations must be unique per LINK-OWNER
-    if conn.isAdmin():
-        for l in old_links:
-            ann = l.child
-            unique_key = "%s_%s" % (l.details.owner.id.val, ann.id.val)
-            links_dict[unique_key] = l
-    # Otherwise, annotations must simply be unique
-    else:
-        for l in old_links:
-            ann = l.child
-            links_dict[ann.id.val] = l
+    # Remove duplicate annotations according to get_key(l)
+    for l in old_links:
+        links_dict[get_key(l, conn.isAdmin())] = l
+
     old_links = links_dict.values()
+
+    # Find existing links on Well so we don't try to duplicate them
+    existing_well_links = list(conn.getAnnotationLinks('Well', [well.id],
+                                                       ns=ns, params=params))
+    existing_well_keys = [get_key(l) for l in existing_well_links]
 
     new_links = []
     for l in old_links:
-        log("    Annotation:", l.child.id.val, l.child.__class__.__name__)
+        if get_key(l) in existing_well_keys:
+            continue
+        log("    Annotation: %s %s" % (l.child.id.val,
+                                       l.child.__class__.__name__))
         link = WellAnnotationLinkI()
         link.parent = WellI(well.id, False)
         link.child = l.child
+        # If Admin, the new link Owner is same as old link Owner
         if conn.isAdmin():
             ownerId = l.details.owner.id.val
             link.details.owner = ExperimenterI(ownerId, False)
@@ -88,17 +103,17 @@ def move_well_annotations(conn, well, ann_type, remove_anns, ns):
     try:
         conn.getUpdateService().saveArray(new_links)
     except Exception, ex:
-        log("Failed to create links: ", ex.message)
+        log("Failed to create links: %s" % ex.message)
         return 0
 
     if remove_anns:
-        log("Deleting ImageAnnotation links...", link_ids)
+        log("Deleting ImageAnnotation links... %s" % link_ids)
         try:
             for link_id in link_ids:
                 to_delete = ImageAnnotationLinkI(link_id, False)
                 conn.getUpdateService().deleteObject(to_delete)
         except Exception, ex:
-            log("Failed to delete links: ", ex.message)
+            log("Failed to delete links: %s" % ex.message)
     return len(new_links)
 
 
@@ -134,7 +149,7 @@ def move_annotations(conn, script_params):
         elif dtype == 'Screen':
             for screen in objects:
                 plates.extend(list(screen.listChildren()))
-        log("Found Plates:", plates)
+        log("Found Plates: %s" % [p.id for p in plates])
         for plate in plates:
             for well in plate.listChildren():
                 ann_count = move_well_annotations(conn, well, filter_type,
@@ -153,7 +168,11 @@ def run_script():
 
     client = scripts.client(
         'Move_Annotations.py',
-        """Move Annotations from SPW Images to their parent Wells.""",
+        """
+For Screen/Plate/Well data, this script moves your Annotations from Images to
+their parent Wells. If you are an Admin, this will also move annotations that
+other users have added, creating links that belong to the same users.
+    """,
 
         scripts.String(
             "Data_Type", optional=False, grouping="1",
--- a/omero/util_scripts/Move_Annotations.py
+++ b/omero/util_scripts/Move_Annotations.py
@@ -21,7 +21,10 @@
 
 import omero.scripts as scripts
 from omero.gateway import BlitzGateway
-from omero.model import WellAnnotationLinkI, WellI, ImageAnnotationLinkI
+from omero.model import ExperimenterI, \
+                        ImageAnnotationLinkI, \
+                        WellAnnotationLinkI, \
+                        WellI
 from omero.rtypes import rstring, rlong
 from omero.constants.metadata import NSINSIGHTRATING
 
@@ -35,9 +38,9 @@ ANN_TYPES = {
 }
 
 
-def log(text):
-    """Handles logging statements in a single place."""
-    print text
+def log(*args):
+    """Handle logging statements in a single place."""
+    print args
 
 
 def move_well_annotations(conn, well, ann_type, remove_anns, ns):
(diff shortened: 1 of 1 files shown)


omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+    'Rating': 'LongAnnotationI',
+    'Key-Value': 'MapAnnotationI'
+}
+
+
+def move_well_annotations(conn, well, ann_type, remove_anns, ns):
+    """Move annotations from Images in this Well onto the Well itself."""
+    print "Processing Well:", well.id, well.getWellPos()
+    iids = [wellSample.getImage().id for wellSample in well.listChildren()]
+    print "  Image IDs:", iids
```
jburel: I have remove ``print`` from all other scripts.
Could you remove it

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+    # Remove duplicate annotations
+    links_dict = {}
+    for l in old_links:
+        ann = l.child
+        links_dict[ann.id.val] = l
+    old_links = links_dict.values()
+
+    new_links = []
+    for l in old_links:
+        print "    Annotation:", l.child.id.val, l.child.__class__.__name__
```
jburel: same

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+        link = WellAnnotationLinkI()
+        link.parent = WellI(well.id, False)
+        link.child = l.child
+        new_links.append(link)
+    try:
+        conn.getUpdateService().saveArray(new_links)
+    except Exception, ex:
+        print "Failed to create links: ", ex.message
```
jburel: same. Log error instead

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+        return 0
+
+    if remove_anns:
+        print "Deleting ImageAnnotation links...", link_ids
```
jburel: same

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+        try:
+            for link_id in link_ids:
+                to_delete = ImageAnnotationLinkI(link_id, False)
+                conn.getUpdateService().deleteObject(to_delete)
+        except Exception, ex:
+            print "Failed to delete links: ", ex.message
```
jburel: same

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+            ann_count = move_well_annotations(conn, well, filter_type,
+                                              remove_anns, ns)
+            ann_total += ann_count
+    else:
+        if dtype == 'Plate':
+            plates = list(objects)
+        elif dtype == 'Screen':
+            for screen in objects:
+                plates.extend(list(screen.listChildren()))
+        print "Found Plates:", plates
```
jburel: same

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+        for plate in plates:
+            for well in plate.listChildren():
+                ann_count = move_well_annotations(conn, well, filter_type,
+                                                  remove_anns, ns)
+                ann_total += ann_count
+
+    return ann_total
+
+
+def runScript():
```
jburel: run_script to match the formatting used in the other script

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+    """The main entry point of the script."""
+    data_types = [rstring('Screen'), rstring('Plate'), rstring('Well')]
+
+    ann_types = [rstring('All')]
+    ann_types.extend([rstring(k) for k in ANN_TYPES.keys()])
+
+    client = scripts.client(
+        'Move_Annotations.py',
+        """Move Annotations from SPW Images to their parent Wells.""",
```
jburel: I won't use SPW, better to be explicit

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+
+        scripts.String(
+            "Data_Type", optional=False, grouping="1",
+            description="The data you want to work with.", values=data_types,
+            default="Plate"),
+
+        scripts.List(
+            "IDs", optional=False, grouping="2",
+            description="List of Screen IDs or Plate IDs").ofType(rlong(0)),
```
jburel: ``well`` is supported as a data_type, this need to be reflected in the description of the ids param

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+
+        scripts.String(
+            "Annotation_Type", grouping="3",
+            description="Move All annotations OR just one type of annotation"
+            " Z below.", values=ann_types, default='All'),
```
jburel: What's ``Z below``?

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+        authors=["William Moore", "OME Team"],
+        institutions=["University of Dundee"],
+        contact="ome-users@lists.openmicroscopy.org.uk",
+    )
+
+    try:
+        conn = BlitzGateway(client_obj=client)
+
+        script_params = client.getInputs(unwrap=True)
+        print script_params
```
jburel: same please remove

jburel:
Since this will be an official test, an integration test must be added to https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/test/integration/scriptstest/test_util_scripts.py

jburel:
``
./omero/util_scripts/Move_Annotations.py:49:29: E128 continuation line under-indented for visual indent
./omero/util_scripts/Move_Annotations.py:123:5: N802 function name should be lowercase
./omero/util_scripts/Move_Annotations.py:184:1: E305 expected 2 blank lines after class or function definition, found 1
``

jburel:
note that on this repo "flake8" is more "aggressive" than on the openmicroscopy repo
i.e. ``flake8 pycodestyle pep8-naming`` are installed

mtbc:
Would it be possible to avoid attempting to link when such a link already exists? Might be especially useful for trying to re-run a script after deleting old links failed.

mtbc:
If an admin runs the script would it be worth trying to preserve the ownership of the previous link?

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,185 @@
...
+    except Exception, ex:
+        print "Failed to create links: ", ex.message
+        return 0
+
+    if remove_anns:
+        print "Deleting ImageAnnotation links...", link_ids
+        try:
+            for link_id in link_ids:
+                to_delete = ImageAnnotationLinkI(link_id, False)
+                conn.getUpdateService().deleteObject(to_delete)
```
mtbc: could delete in bulk?

jburel:
@mtbc: I think we have to preserve the ownership of the links

omero/util_scripts/Move_Annotations.py
will-moore: Ahhhhh! The prints for scripts provide the log for the user. If you don't have any prints then they won't know what's gone wrong or won't have any "info" on what the script has done. They are essential.
At least where I've used a ```log()``` function to do the printing, it's only a case of adding back the print statement there, but for scripts with individual print statements this will be a fair bit harder.

jburel: for now I will follow what has been done in the export scripts i.e. log and do a print at the end if required (that could be turned into an option)
and many of those prints are not necessary
they are helpful when debugging.
We will have to review that part in another round (analyis scripts initialize a logger that is not used)

joshmoore:
NB: This seems like code ripe for wrapping into a server-side method: "relink". cc: @mtbc

mtbc:
Certainly feasible. If it might be done enough in general to warrant server-side support, might be worth a design issue to discuss exactly what it should be able to do and what use cases should be supported.

mtbc:
I'm curious about users in read-only groups running this on others' data: what happens?

Also, can I make a plate from someone else's dataset in a read-only group? What happens then when the annotations are being moved from a same-owner link to a different-owner link (whether by the plate owner or the image owner)?

mtbc:
Would a dry-run mode be useful?

will-moore:
After discussion with @pwalczysko and @mtbc, we decided to only move links that you own (unless you are an Admin).
This simplifies lots of questions about whether to delete another owner's links that you have moved, how to distinguish between new links that were generated from your own links vv generated from someone else's links, whether the script should behave differently in read-write vv read-ann group etc.

Also, we now check for existing links, so we should avoid ValidationExceptions (e.g. if you run the script twice without removing annotations).

pwalczysko:
``Remove Annotations`` --> ``Remove Annotations from Images`` (first thought this would mean completely remove all annotations and do nothing else (i.e. do not add them to wells). If you add ``...from Images`` than it is clearer you are not talking about complete wipeout of the annotations.

pwalczysko:
RFE: put some helpful icons into activities which would lead you to the well with the freshly attached annotations (just like for the other scripts). 
screen shot 2017-03-09 at 12 29 51

pwalczysko:
I don't think that this box should be ticked by default.
screen shot 2017-03-09 at 13 24 13

Especially considering that you can re-run the script a second time with the same settings, only ticking the box and the annotations will still be removed from the images.

Thus, the user will be hopefully inclined to run the script the first time, not deleting anything, when they are happy and have checked, they might want to run the script the second time with the box ticked.

will-moore:
@jburel What do you think to @pwalczysko's suggestion that "Remove Annotations" should not be checked by default?

jburel:
I will agree with @pwalczysko 
since we have moving towards "dry-run" for all actions, I will say not ticked by default too

pwalczysko:
Tested up till now:
 
  - work on my own data ony
           - checked all types of annotations (Tables are not an annotation in true sense, not tested as there is no object in the back end to manipulate)
           - checked the checkbox "Remove annotations"
           - checked filtering for Map Annotations (namespace and Type)
           - checked selecting Well and Plate (it does all wells in that Plate)
           - ToDo: Check selecting Screen with multiple Plates

   - ToDo: work as an admin on other's data (expected behaviour is: all the new links belong to the owners of the corresponding original links)
   - ToDo: work on your own data, but with somebody else annotating them as well (expected behaviour is: only your own links will be transferred, the other user's links will be left as they are)

dominikl:
Looks good, tested (locally) with Screen, Plate and Well <-> Tags, Map, File, Rating, Comment. Only own annotations get moved, unless run as admin (in which case original ownership is preserved) 👍

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,234 @@
...
+from omero.rtypes import rstring, rlong, robject
+from omero.constants.metadata import NSINSIGHTRATING
+
+
+ANN_TYPES = {
+    'Tag': 'TagAnnotationI',
+    'File': 'FileAnnotationI',
+    'Comment': 'CommentAnnotationI',
+    'Rating': 'LongAnnotationI',
+    'Key-Value': 'MapAnnotationI'
```
mtbc, line 38: This subset is basically those that the web UI will usefully display for wells?

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,234 @@
...
+            description="The data you want to work with.", values=data_types,
+            default="Plate"),
+
+        scripts.List(
+            "IDs", optional=False, grouping="2",
+            description="List of Screen, Plate or Well IDs").ofType(rlong(0)),
+
+        scripts.String(
+            "Annotation_Type", grouping="3",
+            description="Move All annotations OR just one type of annotation",
```
mtbc: why is "All" capitalized?

pwalczysko:
Retest in UI went fine, finished all the ToDo's from https://github.com/ome/scripts/pull/134#issuecomment-285363851 as well as checked the new label on the checkbox ``Remove....`` and the new navigation tool in the activities, which works fine.

Ready to merge after the questions of @mtbc are answered.

omero/util_scripts/Move_Annotations.py
will-moore: Yes. My assumption is that users who have added annotations via their own scripts / clients will either have given them a useful namespace that can be used to select them in the Move_Annotations script OR they can also write their own script (or update this one) to customise the task.
Although I could add extra types here, I think this adds extra complexity for users and more testing for us which is probably not justified at this stage.

mtbc: Fair enough, thank you.

omero/util_scripts/Move_Annotations.py
will-moore: Ooops - getting a touch of Title Case here. https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98#.h1s476nc6. I'll fix...

will-moore: I guess it's because "All" is an option in the dropdown list

will-moore:
NB: I still need to finish the integration test for this script, and also add support for linking to Wells from the 'Browse' button in the Activities panel.

jburel:
one RFE: Support for plate acquisition in the type

omero/util_scripts/Move_Annotations.py
```diff
@@ -0,0 +1,234 @@
...
+    except Exception, ex:
+        log("Failed to create links: %s" % ex.message)
+        return 0
+
+    if remove_anns:
+        log("Deleting ImageAnnotation links... %s" % link_ids)
+        try:
+            for link_id in link_ids:
+                to_delete = ImageAnnotationLinkI(link_id, False)
+                conn.getUpdateService().deleteObject(to_delete)
```
jburel: you should not use the deprecated method, use delete2 instead

jburel:
``./omero/util_scripts/Move_Annotations.py:24:1: F401 'omero.model.ImageAnnotationLinkI' imported but unused``

jburel:
https://trello.com/c/TNJk38iN/36-move-annotation-rfe to capture suggestions

jburel:
Thanks for making the changes.
2 failures with the test currently

PR #135, opened 2017-06-22
label_rotate_fix

will-moore:
See https://trello.com/c/rE4VXlX1/33-figure-scripts-rotate-labels-bug
Fixes bug in rotation of labels for figure scripts, previously fixed in OMERO.figure https://github.com/ome/omero-figure/pull/140/files and reported at https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8303

To test, create a Split_View_Figure with the script, and check the vertical labels.
Before and after screenshot:

screen shot 2017-06-22 at 11 59 42

### DIFF
--- a/omero/figure_scripts/Movie_Figure.py
+++ b/omero/figure_scripts/Movie_Figure.py
@@ -334,7 +334,7 @@ def add_left_labels(panel_canvas, image_labels, row_index, width, spacer):
     # add text to rows
     # want it to be vertical. Rotate and paste the text canvas from above
     if image_labels:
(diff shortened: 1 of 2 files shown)


jburel:
I have tried on eel and nightshade, the label is correctly displayed with and without the PR 
eel has Pillow==2.7.0 installed.  
So the change is not "break" older version
I will need to test it locally where I have a newer version of Pillow

jburel:
Tested with Pillow==3.4.2
Label is not correctly rotated without the changes
Including this change for the coming 5.3.3 release

PR #136, opened 2017-10-18
Test in travis

jburel:
With this PR, all scripts are tested in travis instead of OMERO.
This should prevent the failure we have in main CI (movie test is off) or devspace CI (driver for that investigation)

The file ``script.py`` has been copied from OMERO but this should be available from ``testlib``
A PR will be required in OMERO in order to change that.

The test for make_movie is marked as broken due to a codec issue.
see https://trello.com/c/AlN5hp6g/144-make-movie-tests-failures

Using another encoder will be a good idea but it is not a top priority at the moment.
I am not planning to do so in this PR.

cc @sbesson @joshmoore 

Such strategy could be used too in https://github.com/openmicroscopy/omero-example-scripts


joshmoore:
Locally passing with:

```
==================== 26 passed, 1 xfailed in 248.56 seconds ====================
```

Let's see what travis says!

joshmoore:
Kaching

joshmoore:
Merging. As per https://trello.com/c/65fh1AUq/48-remove-pytest-command-and-clipy, I'm going to delay trying to remove `cli.py` and `PytestCommand` until we're closer to the 5.4.2 release.

jburel:
@joshmoore 2dff8d7 and 3674b20 will need to be extracted  since we now have the same problem on the newly created worfklows branch and on https://github.com/ome/scripts/pull/137

PR #143, opened 2018-08-29
Combine Images setPlane

will-moore:
https://trello.com/c/xxnprbWz/35-combineimages-use-setplane

Simply use ```rawPixelStore.setPlane()``` instead of ```setRow()``` to improve speed of CombineImages.py script.

To test:
 - check that script still working as normal.
 - Travis should also be green (integration test passing)

### DIFF
--- a/omero/util_scripts/Combine_Images.py
+++ b/omero/util_scripts/Combine_Images.py
@@ -367,8 +367,8 @@ def make_single_image(services, parameter_map, image_ids, dataset, colour_map):
                     pixel_sizes['y'].append(pixels.getPhysicalSizeY())
                 else:
                     plane_2d = zeros((size_y, size_x))
-                script_utils.upload_plane_by_row(
-                    raw_pixel_store_upload, plane_2d, the_z, the_c, the_t)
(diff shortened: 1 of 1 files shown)


jburel:
Tested on same image on latest and merge
This is certainly an improvement
Ready to merge

joshmoore:
Quick question: what was the largest size image that this was tested with? At what size would we expect this to have issues?

will-moore:
Both before and after this PR, the script fails with Big images (internal exception ```Non-tile based writing unsupported.```) and works OK with images just under 3k x 3k. Tested with ```2789 x 2865```.

joshmoore:
Thanks for checking that, @will-moore. Down the road I could see one or more of these as defensive programming items in the scripts:

 * Doc string saying, "not big image safe"
 * Separate subdirectory for big/non-big image-compatible scripts
 * A script_utils or gateway method which asserts "not is_big_image"
 * A keyword on scripts for whether or not they work with big images

But considering this didn't work before or after at > 3k x 3k, merging.

PR #146, opened 2019-03-18
Add 'Images_Per_Well' support to Dataset_To_Plate

will-moore:
See recent queries about adding multiple Images per Well: https://www.openmicroscopy.org/community/viewtopic.php?f=4&t=8711&p=20693,
and https://www.openmicroscopy.org/community/viewtopic.php?f=6&t=8706

This PR updates the ```Dataset_To_Plate``` script to support multiple Images per Well.

To test:
 - The script should work "as normal" but the Well Per Image parameter can be used to add multiple Images per Well.
 - If there aren't enough images to fill the last Well, it will simply have fewer WellSamples than the other Wells.

### DIFF
--- a/omero/util_scripts/Dataset_To_Plate.py
+++ b/omero/util_scripts/Dataset_To_Plate.py
@@ -39,9 +39,9 @@ import omero
 from omero.rtypes import rint, rlong, rstring, robject, unwrap
 
 
-def add_image_to_plate(conn, image, plate_id, column, row, remove_from=None):
+def add_images_to_plate(conn, images, plate_id, column, row, remove_from=None):
     """
-    Add the Image to a Plate, creating a new well at the specified column and
+    Add the Images to a Plate, creating a new well at the specified column and
     row
     NB - This will fail if there is already a well at that point
     """
@@ -51,22 +51,23 @@ def add_image_to_plate(conn, image, plate_id, column, row, remove_from=None):
     well.plate = omero.model.PlateI(plate_id, False)
     well.column = rint(column)
     well.row = rint(row)
-    well = update_service.saveAndReturnObject(well)
 
     try:
-        ws = omero.model.WellSampleI()
-        ws.image = omero.model.ImageI(image.id, False)
-        ws.well = well
-        well.addWellSample(ws)
-        update_service.saveObject(ws)
+        for image in images:
+            ws = omero.model.WellSampleI()
+            ws.image = omero.model.ImageI(image.id, False)
+            ws.well = well
+            well.addWellSample(ws)
+        update_service.saveObject(well)
     except Exception:
         return False
 
     # remove from Datast
-    if remove_from is not None:
-        links = list(image.getParentLinks(remove_from.id))
-        for l in links:
-            conn.deleteObjectDirect(l._obj)
+    for image in images:
+        if remove_from is not None:
+            links = list(image.getParentLinks(remove_from.id))
+            for l in links:
+                conn.deleteObjectDirect(l._obj)
     return True
 
 
@@ -115,10 +116,17 @@ def dataset_to_plate(conn, script_params, dataset_id, screen):
     if remove_dataset:
         remove_from = dataset
 
-    for image in images:
-        added_count = add_image_to_plate(conn, image,
-                                         plate.getId().getValue(),
-                                         col, row, remove_from)
+    images_per_well = script_params["Images_Per_Well"]
+    image_index = 0
+
+    while image_index < len(images):
+
+        well_images = images[image_index: image_index + images_per_well]
+        added_count = add_images_to_plate(conn, well_images,
+                                          plate.getId().getValue(),
+                                          col, row, remove_from)
+        image_index += images_per_well
+
         # update row and column index
         if first_axis_is_row:
             row += 1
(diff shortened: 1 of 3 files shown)


omero/util_scripts/Dataset_To_Plate.py
```diff
@@ -115,10 +116,17 @@ def dataset_to_plate(conn, script_params, dataset_id, screen):
...
-    for image in images:
-        added_count = add_image_to_plate(conn, image,
-                                         plate.getId().getValue(),
-                                         col, row, remove_from)
+    images_per_well = script_params["Images_Per_Well"]
+    image_index = 0
+
+    while image_index < len(images):
+
+        well_images = images[image_index: image_index + images_per_well]
```
sbesson, line 124: So this means only the following orders should be supported: `[well sample, row, column]` or `[well sample, column, row]`. Does that match the use case reported in the original thread or is it possible we might have to support more combinations?

sbesson (commented):
One question about the expected order of the images. This feature is probably the occasion to start adding integration test for the `Dataset_To_Plate` scripts.

will-moore:
@sbesson We have a test for this script at https://github.com/ome/scripts/blob/6e9ca9bc26182a8f39448e413c74ecf93c268850/test/integration/test_util_scripts.py#L130
but it would need a fair bit more work to support the matrix of various input values for all the script parameters. Maybe I'll just add the new ```Images_Per_Well``` parameter.

sbesson:
Thanks @will-moore. Agreed that covering the whole set of input values is outside the scope of this change and :+1: for focusing on the newly added parameter.

will-moore:
Added test to test the Images_Per_Well parameter (and a couple of others)

jburel:
```
./test/integration/test_util_scripts.py:145:9: F841 local variable 'n' is assigned to but never used
./test/integration/test_util_scripts.py:196:5: E303 too many blank lines (2)
```

test/integration/test_util_scripts.py
```diff
@@ -136,45 +141,55 @@ def test_dataset_to_plate(self):
...
-        for i in range(n):
+        for i in image_names:
             # x,y,z,c,t
-            image = self.create_test_image(100, 100, 1, 1, 1, session)
+            image = self.create_test_image(100, 100, 1, 1, 1, session, name=i)
             self.link(dataset, image, client=client)
             image_ids.append(image.id.val)
 
-        # run the script twice. First with all args...
+        # We run the script twice (parametrize). First with minimum args...
```
sbesson: I was a bit confused by this comment. For each parametrized test, we only run the script right, is that correct? Only the parameters passed to each execution depend on the setup.

omero/util_scripts/Dataset_To_Plate.py
```diff
@@ -39,9 +39,9 @@
 from omero.rtypes import rint, rlong, rstring, robject, unwrap
 
 
-def add_image_to_plate(conn, image, plate_id, column, row, remove_from=None):
+def add_images_to_plate(conn, images, plate_id, column, row, remove_from=None):
```
sbesson, line 42: I assume as this is targetted for OMERO 5.5, changing this API is fine although if we wanted to maintain compatibility for anyone who consumed this API, it should be straightforward to proxy the old API/signature as:
 

```
def add_image_to_plate(conn, image, plate_id, column, row, remove_from=None):
    return add_images_to_plate(conn, [image], plate_id, column, row, remove_from=remove_from)
```

sbesson (commented):
Newly integration test parameter cover the semantics of the script extension. And the explicit usage of `image_names` makes them easier to read and understand the logic.

Barring one comment about the dropped function, this looks good to merge from a code and test review perspective as Travis is green although someone might want to have a quick functional review as well.

omero/util_scripts/Dataset_To_Plate.py
will-moore: I don't think that internal script methods will be used externally. Better to keep this simple since the main function of these scripts is as examples and an unused ```add_image_to_plate()``` is confusing.

test/integration/test_util_scripts.py
will-moore: Simplified comment.

omero/util_scripts/Dataset_To_Plate.py
```diff
@@ -51,22 +51,23 @@ def add_image_to_plate(conn, image, plate_id, column, row, remove_from=None):
...
-        ws.image = omero.model.ImageI(image.id, False)
-        ws.well = well
-        well.addWellSample(ws)
-        update_service.saveObject(ws)
+        for image in images:
+            ws = omero.model.WellSampleI()
+            ws.image = omero.model.ImageI(image.id, False)
+            ws.well = well
+            well.addWellSample(ws)
+        update_service.saveObject(well)
```
joshmoore, line 61: Mostly out of curiosity: why did you change from ws to well?

mtbc (approved):

omero/util_scripts/Dataset_To_Plate.py
will-moore: I use ```ws``` for each WellSample, but now we have many new ```ws``` and I want to save them all in one shot, I can do that by saving the ```well```.

joshmoore:
Ah of course, thanks. That’s what I get for reviewing with whitespace off

PR #147, opened 2019-04-11
batch export ROI coordinates

will-moore:
Noticed during the Dundee workshop - the Batch_ROI_Export script didn't include shape coordinates, lengths or areas. This adds them to the exported CSV, similar to export from iviewer: https://github.com/ome/omero-iviewer/pull/257

To test:
 - Draw all types of Shape on one or more images.
 - Compare export of ROIs csv from iviewer and the output of the Batch_ROI_Export script.
 - Test on images with various pixel sizes or None. If *any* images have NO pixel size set, we use 'pixels' for length and area for all images (since we can't convert).

NB: We use the same pixel size units for the length and area of shapes from all images (since the column has a single unit in header). If some images have very different pixel size units from other images, we could get some very big or small numbers in the CSV, but we can probably assume they are similar if being exported together.

NB: The length and area are converted to length units (e.g. microns) but the shape coordinates are not. They are given in pixel coordinates, including e.g. the Rectangle Width and Height. This is the same behaviour as for iviewer ROI export. I think it makes sense and hope this will be understood by users.

### DIFF
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -193,7 +193,8 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
     if isinstance(shape, PolygonI):
         # https://www.mathopenref.com/coordpolygonarea.html
         coords = shape.getPoints().getValue().split(" ")
-        coords = [[float(x) for x in coord.split(",")] for coord in coords]
+        coords = [[float(x.strip(", ")) for x in coord.split(",")]
+                  for coord in coords]
         total = 0
         for c in range(len(coords)):
             coord = coords[c]
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -2,7 +2,7 @@
 # -*- coding: utf-8 -*-
 
 # -----------------------------------------------------------------------------
-#   Copyright (C) 2018 University of Dundee. All rights reserved.
+#   Copyright (C) 2018-2019 University of Dundee. All rights reserved.
 
 #   This program is free software; you can redistribute it and/or modify
 #   it under the terms of the GNU General Public License as published by
@@ -25,6 +25,9 @@
 import omero.scripts as scripts
 from omero.gateway import BlitzGateway
 from omero.rtypes import unwrap, rstring, rlong, robject
+from omero.model import RectangleI, EllipseI, LineI, PolygonI, PolylineI, \
+    MaskI, LabelI, PointI
+from math import sqrt, pi
 
 DEFAULT_FILE_NAME = "roi_intensities.csv"
 
@@ -34,9 +37,14 @@ def log(data):
     print data
 
 
-def get_export_data(conn, script_params, image):
+def get_export_data(conn, script_params, image, units):
     """Get pixel data for shapes on image and returns list of dicts."""
     log("Image ID %s..." % image.id)
+
+    # Get pixel size in SAME units for all images
+    pixel_size_x = image.getPixelSizeX(units=units).getValue()
+    pixel_size_y = image.getPixelSizeY(units=units).getValue()
+
     roi_service = conn.getRoiService()
     all_planes = script_params["Export_All_Planes"]
     size_c = image.getSizeC()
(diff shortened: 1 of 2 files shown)


pwalczysko:
The resulting csv is still named "roi-intensities". Maybe "Batch ROI export" ?

pwalczysko:
comparing an ellipse area between

   - iviewer (1.291 mm)
   - this script (1.29338894144 mm)

I see two things
   - iviewer rounds up more (not sure if this is good or bad)
   - rounding cannot explain the discrepancy though, the number is clearly different 

user-3
http://web-dev-merge.openmicroscopy.org/webclient/?show=image-70106

On the other hand, line and polyline shows a bit different behaviour (than the ellipse described above) - all differences can be explained by rounding errors.

pwalczysko:
user-3 http://web-dev-merge.openmicroscopy.org/webclient/?show=image-162101

the order of ROIs in the CSV is different between iviewer and the script here.
iviewer is taking the order of the ROIs from the ROI table, seems to me.

Batch ROI export is having its own ordering of ROIs, which I do not see at the moment the logic of.

pwalczysko:
Tested up till now

Calibration

   - none
   - milimeters
   - micrometers

Some comments made

will-moore:
Those commits fix the default name of the csv to ```Batch_ROI_Export.csv``` and sort the ROIs by ID (which I think is what iviewer is doing now).
But, I don't see a difference in ordering for the image above user-3 http://web-dev-merge.openmicroscopy.org/webclient/?show=image-162101

Screen Shot 2019-04-18 at 11 08 34

pwalczysko:
I think we should also revise the allocation of this script under "Export". Note: This script is not exporting ROIs. Unlike the Batch image export, which is exporting images.

I think this script belongs under Analysis.

pwalczysko:
>But, I don't see a difference in ordering for the image above user-3 http://web-dev-merge.openmicroscopy.org/webclient/?show=image-162101

Yes, I cannot repeat that myself now, but I guess it is good to have the ROIs sorted.

will-moore:
The difference in Ellipse area between iviewer and Export script is an iviewer issue fixed in https://github.com/ome/omero-iviewer/pull/263

will-moore:
I think this is still an "Export" script. Image export is different from downloading originals. It's exporting *as" png etc. Here, we are exporting ROIs *as* csv. Also, the script is called ```Batch_ROI_Export.py``` and the equivalent functionality in iviewer itself is ```ROIs > Export as Table```.

To test last commits:
 - Check that default csv name is ```Batch_ROI_Export.csv```
 - Check that ROIs are ordered by ID
 - Check that Ellipse area is same in iviewer (see https://github.com/ome/omero-iviewer/pull/263)

pwalczysko:
>I think this is still an "Export" script. Image export is different from downloading originals. It's exporting *as" png etc. Here, we are exporting ROIs as csv. Also, the script is called Batch_ROI_Export.py and the equivalent functionality in iviewer itself is ROIs > Export as Table.

Yes, understood, but these are all internal naming decisions. Camparison with other softwares shows that such functionality is always under Measure, Analyse etc. We are exporting values, measurements, coordinates.  Saying that we are "exporting ROIs" is too much of a shortcut, and, more importantly, it does not ring a bell with what the user heard elsewhere IMHO

pwalczysko:
testing on eel user-3

  - naming of the produced csv is as expectted
  - order of ROIs is as expected
  - ellipse - see below

Ellipse:
Image  http://web-dev-merge.openmicroscopy.org/webclient/?show=image-21551 user-3
   - in iviewer 37.839
   - this script 37.9016759

I probably do not understand the rounding - not sure how would I round the numbers to get from the script one to the iviewer one. They seem different to me. Also, still not clear why the decimanl places after the point should be so many in the script, so little in iviewer.

will-moore:
For that image, I see a different Area in iviewer: ```37.902```
For script I see the same: ```37.9016759387```

Screen Shot 2019-04-23 at 13 38 21

I can round the Length and Area in the script to 3 decimal places...

pwalczysko:
>I can round the Length and Area in the script to 3 decimal places...

Actually, this is fine. Maybe lets better stick to the current status - more decimal places in the CSV seem a good idea to me, Excel can easily round it for you later.

yes, the image shows now the same elipse size in iviewer and script https://github.com/ome/scripts/pull/147#issuecomment-485787991

All good here

PR #148, opened 2019-06-20
Populate metadata Datasets and Projects

manics:
Requires the omero-metadata plugin. Otherwise reverts to the old behaviour and includes a deprecation warning in the script description.

Updated 20190703: New functionality requires https://github.com/ome/omero-metadata/pull/19

### DIFF
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -31,20 +31,30 @@ from omero.model import PlateI, ScreenI
 import sys
 
 from omero.util.populate_roi import DownloadingOriginalFileProvider
-from omero.util.populate_metadata import ParsingContext
+try:
+    # Hopefully this will import
+    # https://github.com/ome/omero-metadata/blob/v0.3.1/src/populate_metadata.py
+    from populate_metadata import ParsingContext
+except ImportError:
+    from omero.util.populate_metadata import ParsingContext
+
+
+OBJECT_TYPES = (
+    'Plate',
+    'Screen',
+    'Dataset',
+    'Project',
+)
 
 
 def get_original_file(conn, object_type, object_id, file_ann_id=None):
-    if object_type == "Plate":
-        omero_object = conn.getObject("Plate", int(object_id))
-        if omero_object is None:
-            sys.stderr.write("Error: Plate does not exist.\n")
-            sys.exit(1)
-    else:
-        omero_object = conn.getObject("Screen", int(object_id))
-        if omero_object is None:
-            sys.stderr.write("Error: Screen does not exist.\n")
-            sys.exit(1)
+    if object_type not in OBJECT_TYPES:
+        sys.stderr.write("Error: Invalid objet type: %s.\n" % object_type)
+        sys.exit(1)
+    omero_object = conn.getObject(object_type, int(object_id))
+    if omero_object is None:
+        sys.stderr.write("Error: %s does not exist.\n" % object_type)
+        sys.exit(1)
     file_ann = None
 
     for ann in omero_object.listAnnotations():
@@ -84,7 +94,7 @@ def populate_metadata(client, conn, script_params):
 
 def run_script():
 
-    data_types = [rstring('Plate'), rstring('Screen')]
+    data_types = [rstring(otype) for otype in OBJECT_TYPES]
     client = scripts.client(
         'Populate_Metadata.py',
         """
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -26,7 +26,7 @@ import omero
 from omero.gateway import BlitzGateway
 from omero.rtypes import rstring, rlong
 import omero.scripts as scripts
-from omero.model import PlateI, ScreenI
+import omero.model
 
 import sys
 
@@ -49,7 +49,7 @@ OBJECT_TYPES = (
 
 def get_original_file(conn, object_type, object_id, file_ann_id=None):
     if object_type not in OBJECT_TYPES:
-        sys.stderr.write("Error: Invalid objet type: %s.\n" % object_type)
+        sys.stderr.write("Error: Invalid object type: %s.\n" % object_type)
         sys.exit(1)
     omero_object = conn.getObject(object_type, int(object_id))
     if omero_object is None:
@@ -82,10 +82,8 @@ def populate_metadata(client, conn, script_params):
         conn, data_type, object_id, file_ann_id)
     provider = DownloadingOriginalFileProvider(conn)
     file_handle = provider.get_original_file_data(original_file)
-    if data_type == "Plate":
-        omero_object = PlateI(long(object_id), False)
-    else:
-        omero_object = ScreenI(long(object_id), False)
+    objectI = getattr(omero.model, data_type + 'I')
+    omero_object = objectI(long(object_id), False)
     ctx = ParsingContext(client, omero_object, "")
     ctx.parse_from_handle(file_handle)
     ctx.write_to_omero()
@@ -98,8 +96,8 @@ def run_script():
     client = scripts.client(
         'Populate_Metadata.py',
         """
-    This script processes a csv file, attached to a Screen or Plate,
-    converting it to an OMERO.table, with one row per Well.
+    This script processes a csv file, attached to a container,
+    converting it to an OMERO.table, with one row per Image or Well.
     The table data can then be displayed in the OMERO clients.
     For full details, see
     http://help.openmicroscopy.org/scripts.html#metadata
@@ -107,11 +105,11 @@ def run_script():
         scripts.String(
             "Data_Type", optional=False, grouping="1",
             description="Choose source of images",
-            values=data_types, default="Plate"),
+            values=data_types, default=OBJECT_TYPES[0]),
 
         scripts.List(
             "IDs", optional=False, grouping="2",
-            description="Plate or Screen ID.").ofType(rlong(0)),
+            description="Container ID.").ofType(rlong(0)),
 
         scripts.String(
             "File_Annotation", grouping="3",
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -34,7 +34,7 @@ from omero.util.populate_roi import DownloadingOriginalFileProvider
 try:
     # Hopefully this will import
     # https://github.com/ome/omero-metadata/blob/v0.3.1/src/populate_metadata.py
-    from populate_metadata import ParsingContext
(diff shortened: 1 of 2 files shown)


pwalczysko:
```
  File "./script", line 140, in <module>
    run_script()
  File "./script", line 132, in run_script
    message = populate_metadata(client, conn, script_params)
  File "./script", line 88, in populate_metadata
    ctx.parse_from_handle(file_handle)
AttributeError: 'ParsingContext' object has no attribute 'parse_from_handle'
```

tested locally

pwalczysko:
@manics I have run the ``pip list`` to find the omero-metadata in the venv where my omero.web is installed (localhost). Maybe I should be installing the omero-metadata somewhere else ?

omero/import_scripts/Populate_Metadata.py
```diff
@@ -84,24 +92,24 @@ def populate_metadata(client, conn, script_params):
...
-            values=data_types, default="Plate"),
+            values=data_types, default=OBJECT_TYPES[0]),
 
         scripts.List(
             "IDs", optional=False, grouping="2",
-            description="Plate or Screen ID.").ofType(rlong(0)),
+            description="Container ID.").ofType(rlong(0)),
 
         scripts.String(
             "File_Annotation", grouping="3",
```
will-moore, line 133: Not related to this PR, but since you're making changes, perhaps you could change this to a ```Long``` input instead of a string?

joshmoore:
cc: @emilroz

manics:
@pwalczysko This should be working now

pwalczysko:
First test (locally) on idr0021 passes.


Tested on merge-ci, where the omero-metadata is not installed. The script correctly defaults to just showing the dropdown with Screen, Plate only in this case.
Then run through the scenario test on Screen and Plate and got 

  - empty values for the columns in the UI
  - a message

```
WARNING:omero.util.populate_metadata:PlateColumn is unimplemented
```

Edit: the problem with empty values in the tables when the test scenario is run on Screen https://docs.openmicroscopy.org/internal/testing_scenarios/BulkAnnotations.html reported in this comment above is not a regression - was able to reproduce it precisely with the old script (script as it was prior to this PR) on the same server (merge-ci, user-3)

Edit2: the problem is only on Screens, not on orphaned Plates - again, this is not a regression and has nothing to do with this PR - this PR seems to preserve the functionality of the old script correctly.

pwalczysko:
on merge-ci, after I did ``pip install omero-metadata`` and redeployed the server, I get no warning in the UI about the outdated plugin which is used (as expected) and the dropdown has now also the Project and Dataset items (as expected).

Two occasions of the below error though, both logged in as

user-2

1. https://merge-ci.openmicroscopy.org/web/webclient/?show=project-3059
2. https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-3001

```
  File "./script", line 159, in <module>
    run_script()
  File "./script", line 151, in run_script
    message = populate_metadata(client, conn, script_params)
  File "./script", line 107, in populate_metadata
    ctx.parse_from_handle_stream(data)
  File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python2.7/site-packages/populate_metadata.py", line 964, in parse_from_handle_stream
    self.create_file_annotation(table)
  File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python2.7/site-packages/populate_metadata.py", line 1090, in create_file_annotation
    original_file = table.getOriginalFile()
  File "/home/omero/workspace/OMERO-server/OMERO.server/lib/python/omero_Tables_ice.py", line 1038, in getOriginalFile
    return _M_omero.grid.Table._op_getOriginalFile.invoke(self, ((), _ctx))
Ice.ObjectNotExistException: exception ::Ice::ObjectNotExistException
{
    id = 
    {
        name = Table-0127863d-28d8-4628-9255-15fe91d1f0be
        category = dacc667e-ee69-44d5-bfd7-dfb1f4dec17d
    }
    facet = 
    operation = getOriginalFile
}
```

joshmoore:
This sounds quite similar to https://github.com/openmicroscopy/openmicroscopy/pull/6066

pwalczysko:
When I run the script on Screen with mias/frans plate in it, as per scenario https://docs.openmicroscopy.org/internal/testing_scenarios/BulkAnnotations.html

user-2, https://merge-ci.openmicroscopy.org/web/webclient/?show=screen-3001

I get

```
WARNING:omero.util.populate_metadata:Screen is missing plate: NameOfThePlate
WARNING:omero.util.populate_metadata:PlateColumn is unimplemented
  File "./script", line 159, in <module>
    run_script()
  File "./script", line 151, in run_script
    message = populate_metadata(client, conn, script_params)
  File "./script", line 106, in populate_metadata
    ctx.preprocess_from_handle(data_for_preprocessing)
  File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python2.7/site-packages/populate_metadata.py", line 938, in preprocess_from_handle
    self.preprocess_data(reader)
  File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python2.7/site-packages/populate_metadata.py", line 1028, in preprocess_data
    self.post_process()
  File "/home/omero/workspace/OMERO-server/omero-virtualenv/lib/python2.7/site-packages/populate_metadata.py", line 1154, in post_process
    plate = columns_by_name["plate"].values[i]
IndexError: list index out of range
```

pwalczysko:
@joshmoore I did not detect this problem locally, running the script on a 5.5.0 release server.

joshmoore:
If you're referring to the `getOriginalFile` error, then yes, that's to be expected. If this is so pervasive, then perhaps we have to consider this a breaking change. The line https://github.com/ome/omero-metadata/blob/master/src/populate_metadata.py#L1090 is the risky one. Note: this also points to a leak though so I'm hesitate to say we should do _something_.

pwalczysko:
@joshmoore 

>If you're referring to the getOriginalFile error, then yes, that's to be expected. If this is so pervasive, then perhaps we have to consider this a breaking change. The line https://github.com/ome/omero-metadata/blob/master/src/populate_metadata.py#L1090 is the risky one. Note: this also points to a leak though so I'm hesitate to say we should do something.

Yes, I do refer to the getOriginalFile error. FMPOV, it would be better to have this PR here in and having the user have a way to create a table from a csv than fixing the complex non-closing problematics.

This PR here is blocked atm by the inability to test its performance on merge-ci because of https://github.com/ome/scripts/pull/148#issuecomment-504477060

pwalczysko:
@manics https://github.com/ome/scripts/pull/148#issuecomment-504478246 seems to be a Screen-specific, separate problem which repeats every time when the script is run on a Screen.  But again, running on a Screen did not work in the first place in the old script... nevertheless, with a different output.

will-moore:
As discussed with @pwalczysko the issue at https://github.com/ome/scripts/pull/148#issuecomment-504478246 was the csv ```Plate``` column still had ```NameOfThePlate``` placeholder values. When we replaced those with the actual name of the Plate, this error went away (with the csv attached to the Screen).

pwalczysko:
On outreach server, as expected, the script as edited by this PR works fine. Again, OMERO.parade is having problems ot open the resulting table, but this is fixed by @will-moore in https://github.com/ome/omero-parade/pull/55

omero/import_scripts/Populate_Metadata.py
```diff
@@ -26,25 +26,45 @@
...
+import omero.model
 
 import sys
 
 from omero.util.populate_roi import DownloadingOriginalFileProvider
-from omero.util.populate_metadata import ParsingContext
+try:
+    # Hopefully this will import
+    # https://github.com/ome/omero-metadata/blob/v0.3.1/src/populate_metadata.py
+    from populate_metadata import ParsingContext
```
joshmoore: If this is renamed, how would you suggest doing the detection?

omero/import_scripts/Populate_Metadata.py
```diff
@@ -26,25 +26,45 @@
...
+
+except ImportError:
+    from omero.util.populate_metadata import ParsingContext
+    OBJECT_TYPES = (
+        'Plate',
+        'Screen',
+    )
+    DEPRECATED = """
+
+    Warning: This script is using an outdated metadata plugin.
```
joshmoore, line 54: This is nice wording. Probably worth doing this everywhere.

joshmoore (commented):
Looks good. I've opened https://github.com/ome/omero-metadata/pull/23 to fix the tables issue that @pwalczysko was seeing while testing.

omero/import_scripts/Populate_Metadata.py
manics: Since this is a new feature I don't think we should support multiple breaking versions. If there's an agreement on renaming the module could we do that now and require that as the minimum version for these new features?

pwalczysko:
Opened https://github.com/manics/scripts/pull/1 to fix the problem mentioned in https://github.com/ome/omero-metadata/pull/23#issuecomment-508146327 cc @joshmoore 


Witch https://github.com/manics/scripts/pull/1 and https://github.com/ome/omero-metadata/pull/23 and https://github.com/ome/omero-metadata/pull/19 all works on merge-ci as epected

joshmoore:
I was going to say:
> Merging for an initial build of 5.5.1 that will be deployed to pub-omero before proper release.

but then I noticed https://github.com/ome/scripts/pull/148/commits/b268cda87f1745f7db0bf12f48ace495005517ad

Can anyone explain this change further? Is this not a breaking change?

cc: @manics @pwalczysko @will-moore

will-moore:
That was my suggestion which doesn't change the usage of the script in the UI but simply means that you're not going to get an invalid String when you're really wanting an ID.
I guess it could be considered a breaking change if you're running the script via code. I didn't consider that because the whole point of the script is to provide a UI entry point to the underlying populate_metadata script, so I don't expect any code to be passing a string here, but I guess it's possible. Happy for it to be reverted.

joshmoore:
Reverted. The change certainly makes sense, but if we want to make it, we probably need either a bigger bump or to deprecate `File_Annotation` in favor of `File_Annotation_ID` (or similar). I don't know if we have a clear deprecation mechanism for script properties though.

joshmoore:
Today's build was green. Including this under the assumption that everything still works (to be re-confirmed by @pwalczysko). I'll get a 5.5.1 build together and deployed so that we have some time to test.

pwalczysko:
@joshmoore still works as expected with omero-metadata==0.4.0 on https://latest-ci.openmicroscopy.org/web/webclient/ user-3 - tested both Plate and Project

pwalczysko:
This is good now, and we have some progress. As discussed with @will-moore , it would be great to have this script to create tables with numbers where there are numbers in the csv.
Atm, the script always creates strings.

PR #150, opened 2019-07-17
Script to delete annotations

dominikl:
I would find a script like this very helpful when testing scripts or preparing training workflows. These tasks usually create lots of file attachments, tables, ROIs, etc. Deleting them manually to get back to a clean state is a pain. With this script one can simply select a project/dataset/image/... and clean it from all annotations/rois.

If it's not generally useful, I just close the PR again...

### DIFF
--- a/omero/util_scripts/Strip.py
+++ b/omero/util_scripts/Strip.py
@@ -1,0 +1,155 @@
+import omero.scripts as scripts
+from omero.gateway import BlitzGateway, ImageWrapper, FileAnnotationWrapper, TagAnnotationWrapper, MapAnnotationWrapper, CommentAnnotationWrapper
+from omero.rtypes import rstring, rlong
+
+# Count deleted annotations
+rois_deleted = 0
+annos_deleted = 0
+
+def delete_annotations(conn, object, del_f=False, del_t=False, del_m=False, del_c=False):
+    """Deletes annotations from an object
+
+    Parameters:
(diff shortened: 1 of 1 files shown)


will-moore:
I can see this being useful for us (OME Training team) but I'm not sure that is enough to justify adding to the scripts we ship. But, many of those scripts are really just examples anyway, so I'm OK to add more.
For training, we have https://github.com/ome/training-scripts/blob/v0.6.0/maintenance/scripts/delete_annotations.py and https://github.com/ome/training-scripts/blob/v0.6.0/maintenance/scripts/delete_ROIs.py

dominikl:
Is there a different repository? This whole omero-user-scripts thing seems to be rather historic, isn't it?

jburel:
As @will-moore indicated, training-script repo is the best place for this type of scripts
Are you thinking of https://github.com/ome/omero-user-scripts?

joshmoore:
Not sure if we already having something similar, but one option would be to split this into a discovery tool like `bin/omero annotations list` (similar to FindChildren and FindParents) and then have the ability to pipe that result to `bin/omero delete`.

cc: @mtbc

mtbc:
This is different from just `bin/omero delete Project/Roi:1234` or whatever I guess?

dominikl:
Whatever, I'd just like to have button (/script) in Web to simply reset any analysis results on a project. Thought that might be generally useful. Closed now.

PR #152, opened 2019-08-16
populate metadata links file ann

will-moore:
This change is needed to allow populate_metadata script to work with the File-picker from https://github.com/ome/openmicroscopy/pull/6096
That PR passes the uploaded FileAnnotation ID to populate-metadata script, but the script expects the FileAnnotation to be linked to the selected object.
This PR links the FileAnnotation to the object, which also has the benefit of not 'losing' the new FileAnnotation with it being unlinked.

To test:
 - Create or find a CSV FileAnnotation for populate metadata. E.g. https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-4919 (user-3)
 - Note the FileAnnotation ID - (9987) then *unlink* (don't delete) the Annotation from the Dataset.
 - Run populate_metadata script on the Dataset, entering the File Annotation ID.
 - Check that the populate-metadata works as expected to create an OMERO.table and that the csv becomes linked to the Dataset.
 - If the File_Annotation ID is filled out in the script (and the FileAnnotation is already linked to the object) then script should continue to work as before

### DIFF
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -57,7 +57,22 @@ except ImportError:
     """
 
 
-def get_original_file(conn, object_type, object_id, file_ann_id=None):
+def link_file_ann(conn, object_type, object_id, file_ann_id):
+    """Link File Annotation to the Object, if not already linked."""
+    file_ann = conn.getObject("Annotation", file_ann_id)
+    if file_ann is None:
+        sys.stderr.write("Error: File Annotation not found: %s.\n"
+                         % file_ann_id)
+        sys.exit(1)
+    omero_object = get_object(conn, object_type, object_id)
+    # Check for existing links
+    links = list(conn.getAnnotationLinks(object_type, parent_ids=[object_id],
+                                         ann_ids=[file_ann_id]))
+    if len(links) == 0:
+        omero_object.linkAnnotation(file_ann)
+
+
+def get_object(conn, object_type, object_id):
     if object_type not in OBJECT_TYPES:
         sys.stderr.write("Error: Invalid object type: %s.\n" % object_type)
         sys.exit(1)
@@ -65,6 +80,11 @@ def get_original_file(conn, object_type, object_id, file_ann_id=None):
     if omero_object is None:
         sys.stderr.write("Error: %s does not exist.\n" % object_type)
         sys.exit(1)
+    return omero_object
+
+
+def get_original_file(conn, object_type, object_id, file_ann_id=None):
+    omero_object = get_object(conn, object_type, object_id)
     file_ann = None
 
     for ann in omero_object.listAnnotations():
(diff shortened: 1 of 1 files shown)


jburel:
*  user-1 merge-ci
* project ID 5757
* Create a file using https://github.com/ome/omero-metadata/ and save locally
* Choose the local file
* ``Error: File does not exist.``

will-moore:
It looks like ```merge-ci``` doesn't have the latest version of the script, although it was merged in the last build at https://github.com/snoopycrimecop/scripts/commit/b7de9d77ffc8264001c9fecd2498474e4f094b15

sbesson:
I was surprised by this at first it looks consistent to the introduction of the `--shallow` flag in https://github.com/ome/devspace/commit/9a99bf129029c867904e0abaf1fd90ae2a4751d3.

will-moore:
Does that commit need reverting, or some other fix?

sbesson:
Reverting the commit will only apply to newly deployed devspace environment. In the case of the existing merge-ci devspace, I think you want to modify the default value of `MERGE_COMMAND` in the job configuration and restore the previous behavior unless @jburel remembers a reason not to revert his change.

jburel:
I do not remember why we added the ``shallow`` flag

joshmoore:
For what it's worth, we will very soon no longer have a submodule in `ome/openmicroscopy`.

joshmoore:
Coming back to the content of the PR:

> This change is needed to allow populate_metadata script to work with the File-picker from ome/openmicroscopy#6096

We might need to discuss how this type of coupling should affect the semver of these two components.

will-moore:
Certainly we shouldn't release the change in the script dialog https://github.com/ome/openmicroscopy/pull/6096 before the script changes in this PR.

There are a few places where we have a 'convention' on the naming of script Parameters. E.g. "Data_Type" and "IDs" for data input. These are quite simple convention so we haven't felt the need to define this API anywhere. The "File_Annotation" parameter is also understood by the webclient and can be auto-populated. I assumed File_Annotation would simply use the ID of a File_Annotation, but it has an additional requirement that the Annotation is linked to the Project/Screen for populate_metadata script.

Perhaps we can test and merge this PR first, then consider https://github.com/ome/openmicroscopy/pull/6096

Will that be enough to ensure that user don't get the webclient dialog changes before they get the script changes?

cc @chris-allan

pwalczysko:
We still do not have this script on the merge-ci - judging according to me getting the same result of the test as @jburel above https://github.com/ome/scripts/pull/152/#issuecomment-522608879 - how to improve that ?

pwalczysko:
Uploaded the script here manually onto merge-ci. The functionality works as described.

will-moore:
Closing until #159 is merged to avoid conflicts...

pwalczysko:
All works fine here. Tried the suggested table as well as my own, with double columns -> parade 

The unattached File is found via the ID as expected, and when one is attached there already, the script still runs as expected.

Ready to merge FMPOV

pwalczysko:
Maybe we should test on py3-ci though ? All the tests above were run on merge-ci

will-moore:
Uploaded and tested on py3-ci today. Will list for testing there tomorrow (will need to upload script again first). Use https://py3-ci.openmicroscopy.org/web/webclient/?show=dataset-3120 (user-3).

pwalczysko:
Tested on py3-ci, user-3 https://py3-ci.openmicroscopy.org/web/webclient/?show=dataset-2270

All good. Ready to merge.

joshmoore:
> This change is needed to allow populate_metadata script to work with the File-picker from ome/openmicroscopy#6096

But that PR is excluded. Can you explain?

will-moore:
https://github.com/ome/openmicroscopy/pull/6096 requires this PR, so it is probably excluded until this one is merged, released etc. However, this PR doesn't require that one.

PR #153, opened 2019-09-02
Set proper version in setup.py

joshmoore:
As with https://github.com/ome/omero-py/pull/15, in order to use:

```
python setup.py --version
```

as the basis for build-infra's `foreach-get-version` the setup.py version will need to be kept in sync with `git describe`.


joshmoore:
Merging proactively. Suggestions on how to simplify the synchronization welcome.

PR #155, opened 2019-10-24
py3 update

jburel:
first round for py3

PR #157, opened 2019-11-06
py3 fixes

will-moore:
To test:
 - Check that all scripts are working as normal

### DIFF
--- a/omero/util_scripts/Channel_Offsets.py
+++ b/omero/util_scripts/Channel_Offsets.py
@@ -360,11 +360,13 @@ See http://help.openmicroscopy.org/scripts.html""",
 
     try:
         script_params = client.getInputs(unwrap=True)
+        print(script_params)
 
         # wrap client to use the Blitz Gateway
         conn = BlitzGateway(client_obj=client)
 
         images, dataset, message = process_images(conn, script_params)
+        print(message)
 
         # Return message, new image and new dataset (if applicable) to the
         # client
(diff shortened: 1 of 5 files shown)


joshmoore:
> we avoid importing script_utils

:+1: seems like it can be more and more deprecated in favor of gateway.

omero/util_scripts/Move_Annotations.py
```diff
@@ -41,7 +41,7 @@
 
 def log(text):
     """Handle logging statements in a single place."""
-    print text
+    print(text)
```
jburel: This is already done in https://github.com/ome/scripts/pull/155

jburel: This will conflict

jburel:
I have merged the other PR, so the last commit can be removed @will-moore

omero/util_scripts/Channel_Offsets.py
```diff
@@ -360,11 +360,13 @@ def run_script():
 
     try:
         script_params = client.getInputs(unwrap=True)
+        print(script_params)
```
joshmoore, line 363: Did you intend for these to stay?

joshmoore (commented):
Other than a question about the print statements, this looks good to go.

omero/util_scripts/Channel_Offsets.py
will-moore: Yes, I think it's generally useful to have some logging of the params (and other info) when you ran a script.

PR #163, opened 2019-11-22
Fix omero-scripts packaging

sbesson:
This defines each script directory as package and declares them in packages for inclusion in the source distribution

See https://github.com/ome/openmicroscopy/pull/6181#issuecomment-557520702. This should be the first step towards being able to upload `omero-scripts` source distributions to PyPI for consumption for the OMERO distribution bundles.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -71,6 +71,11 @@ url = "https://github.com/ome/scripts/"
 setup(
     version=version,
     name='omero-scripts',
+    packages=[
+        'omero.analysis_scripts',
+        'omero.export_scripts',
+        'omero.import_scripts',
+        'omero.util_scripts'],
     description="OMERO scripts",
     long_description="OMERO scripts",
     author='The Open Microscopy Team',


joshmoore:
Makes sense. We'll just need to be careful when adding new directories.

joshmoore:
Still looks ok. I understand the only change to be rebasing onto the new release? If so, :+1:

sbesson:
See the tar.gz of https://py3-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-build/ which now includes all scripts.

Proposing to open a PR against ome/openmicroscopy to consume this source distribution artifact rather than the GitHub archive and have it tested by daily CI builds. If tests keep passing, proposed next steps would be: 1- rename this repository as `omero-scripts`, 2- bump the version to `5.6.dev1`, 3- add the infrastructure to release to PyPI, 4- create a pre-release, 5- update `etc/omero.properties` to unpack this artifact like the other Python components

joshmoore:
In general, all sounds fine. The only thing that should be verified is that `pip install omero-scripts` doesn't lead to any weirdness. If it *does*, then we should make that clear on PyPI.

joshmoore (approved):

PR #165, opened 2019-12-17
batch ROI export CSV unicode

### LINKED ISSUE #158: py3 issues
Batch_ROI_Export failing because of unicode um characters at
```
 File "./script", line 335, in run_script
    result = batch_roi_export(conn, script_params)
  File "./script", line 280, in batch_roi_export
    file_ann = write_csv(conn, export_data, script_params, symbol)
  File "./script", line 238, in write_csv
    csv_file.write("\n".join(csv_rows))
UnicodeEncodeError: 'ascii' codec can't encode character '\xb5' in position 64: ordinal not in range(128)
```

Populate_Metadata.py - opening script dialog gives:
```
omero.ValidationException: exception ::omero::ValidationException
{
    serverStackTrace = 
    serverExceptionClass = 
    message = Can't find params for 266058!
No stdout.
stderr is in file 266442:
---------------------------------
  File "./script", line 43, in <module>
    from omero_metadata.populate import ParsingContext
  File "/opt/omero/server/venv3/lib64/python3.6/site-packages/omero_metadata/populate.py", line 87
    Report bugs to ome-devel@lists.openmicroscopy.org.uk""" % (error, cmd, cmd)
                                                          ^
SyntaxError: invalid syntax
```

will-moore:
Fixes #158.
This reverts commit d9119def3237feefe45b7972e024eb159102d98d.
Shouldn't be needed once we have sys.stdout.encoding UTF-8 on
the server.

Requires https://github.com/ome/omero-py/pull/150

To test, export ROIs from an image with pixel size of microns and check that µm is written to CSV columns correctly.

### DIFF
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -228,18 +228,14 @@ def write_csv(conn, export_data, script_params, units_symbol):
         units_symbol = "pixels"
     csv_header = csv_header.replace(",length,", ",length (%s)," % units_symbol)
     csv_header = csv_header.replace(",area,", ",area (%s)," % units_symbol)
-    csv_rows = [csv_header.encode('utf-8')]
+    csv_rows = [csv_header]
 
     for row in export_data:
-        # cells = [("%s" % row.get(name, "")) for name in COLUMN_NAMES]
-        cells = []
-        for name in COLUMN_NAMES:
-            td = row.get(name, '')
-            cells.append(str(td).encode('utf-8'))
-        csv_rows.append(b",".join(cells))
+        cells = [str(row.get(name, "")) for name in COLUMN_NAMES]
+        csv_rows.append(",".join(cells))
 
-    with open(file_name, 'wb') as csv_file:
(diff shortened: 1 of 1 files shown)


jburel:
script not available

will-moore:
Don't know why scripts are not uploaded to the server. But I just uploaded this script manually (from this branch).

jburel:
``area (µm)`` This should be ``area (µm²)``
iviewer works correctly

also the values could be rounded e.g. ``66.95859863900198`` => ``66.959``

will-moore:
This PR is just rolling back the ```.encode()``` that we don't need now. Not intending to add new features.
Also, I don't know how easy it is to write superscript to a text file.
I'm not sure we should be rounding here as the output is not for display but for further analysis
(equivalent to exporting a lossy jpeg instead of tiff for analysis).

jburel:
My point is that iviewer and the script generate both a CSV but one round the result one does not
that could be very confusing for user
I will create an issue for unit display

jburel:
issues #168 and #169 created
Merging the PR

PR #170, opened 2020-02-02
ROI export handle points whitespace

will-moore:
Found this bug when trying Batch_ROI_Export from outreach server, using idr0022 data.
Failed with ```float(x.strip(", ")``` since points string was not in the exact expected format.

I now strip whitespace before split(' ') so don't end up with empty string from trailing whitespace.
Also wrap with try/except in case some other unexpected format - Don't want a single invalid
points list to break the whole script.

To test:
 - Upload script to e.g. outreach server and try Batch_ROI_Export on idr0021 data.

### DIFF
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -194,7 +194,7 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
             coords = [[float(x.strip(", ")) for x in coord.split(",", 1)]
                       for coord in coords]
         except ValueError:
-            print("Invalid Polyline coords:", coords)
+            log("Invalid Polyline coords: %s" % coords)
         else:
             lengths = []
             for i in range(len(coords)-1):
@@ -211,7 +211,7 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
             coords = [[float(x.strip(", ")) for x in coord.split(",", 1)]
                       for coord in coords]
         except ValueError:
-            print("Invalid Polygon coords:", coords)
+            log("Invalid Polygon coords: %s" % coords)
         else:
             total = 0
             for c in range(len(coords)):
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -192,7 +192,7 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
         coords = point_list.strip(" ").split(" ")
         try:
             coords = [[float(x.strip(", ")) for x in coord.split(",", 1)]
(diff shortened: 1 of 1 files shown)


pwalczysko:
Tested on merge-ci user-3 (we have the idr0021 in full with ROIs imported there too).

All works fine, on the dataset which errored on ome-training-4 previously, as well as on the whole idr0021 project.

Ready to merge fmpov

omero/export_scripts/Batch_ROI_Export.py
```diff
@@ -189,28 +189,36 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
...
-            dx = dx if pixel_size_x is None else dx * pixel_size_x
-            dy = dy if pixel_size_y is None else dy * pixel_size_y
-            lengths.append(sqrt((dx * dx) + (dy * dy)))
-        row_data['length'] = sum(lengths)
+        coords = point_list.strip(" ").split(" ")
+        try:
+            coords = [[float(x.strip(", ")) for x in coord.split(",", 1)]
+                      for coord in coords]
+        except ValueError:
+            print("Invalid Polyline coords:", coords)
```
jburel: preferable to use the ``log`` method, 
if we decide to change its implementation it will be much easier.

omero/export_scripts/Batch_ROI_Export.py
```diff
@@ -189,28 +189,36 @@ def add_shape_coords(shape, row_data, pixel_size_x, pixel_size_y):
...
-            coord = coords[c]
-            next_coord = coords[(c + 1) % len(coords)]
-            total += (coord[0] * next_coord[1]) - (next_coord[0] * coord[1])
-        row_data['area'] = abs(0.5 * total)
+        coords = point_list.strip(" ").split(" ")
+        try:
+            coords = [[float(x.strip(", ")) for x in coord.split(",", 1)]
+                      for coord in coords]
+        except ValueError:
+            print("Invalid Polygon coords:", coords)
```
jburel: same

joshmoore:
The idea would be to roll this into the next release, correct? Is there anything else foreseen or shall we call this `5.6.1`?

jburel:
@will-moore might come back from workshop with a fix for a script :-)
in that case it might be included in a 5.6.1
minor comments on this PR that will be good to have it in

jburel:
I could not reproduce the error on workshop using few images from idr21 (images 9539, 9540, 9541) using the older version of the script

jburel:
The handling of the string is quite fragile in a way
This PR handles potential cases we have seen and it does not affect existing shapes

will-moore:
No, I can't find the shapes that broke this script originally.
But you only need to have Points string with trailing whitespace ```"1,2 3,4 "``` to break the script (without this PR).

PR #171, opened 2020-02-14
Update url in populate metadata description

will-moore:
Now we link to https://github.com/ome/omero-metadata/
for how to format the CSV instead of old help site.

### DIFF
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -1,7 +1,7 @@
 # coding=utf-8
 '''
 -----------------------------------------------------------------------------
-  Copyright (C) 2014 Glencoe Software, Inc. All rights reserved.
+  Copyright (C) 2014-2020 Glencoe Software, Inc. All rights reserved.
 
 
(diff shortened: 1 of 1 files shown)


omero/import_scripts/Populate_Metadata.py
```diff
@@ -136,11 +136,11 @@ def run_script():
...
         'Populate_Metadata.py',
         """
-    This script processes a csv file, attached to a container,
+    This script processes a CSV file, attached to a container,
     converting it to an OMERO.table, with one row per Image or Well.
     The table data can then be displayed in the OMERO clients.
-    For full details, see
-    http://help.openmicroscopy.org/scripts.html#metadata
+    For full details of the supported CSV format, see
+    https://github.com/ome/omero-metadata/
```
manics, line 143:
```suggestion
    https://github.com/ome/omero-metadata/blob/master/README.rst
```
Might be better to link directly to the README? Not essential though

will-moore: I think I prefer the shorter URL. It might be more stable too.

PR #173, opened 2020-03-07
ROI export with Well ID

will-moore:
When exporting ROIs to csv from SPW data, it is useful to include Well ID (and Well row, column and label).
This also adds the option to NOT export the Points coordinates string, which can cause a big increase in the size of the csv file.

This was needed for OMERO.parade prototype demo at https://youtu.be/FyjGhZxx6es?t=899

To test:
 - Export ROIs for a Screen or Plate.
 - Check that the CSV has columns for Well ID, row, column and label.
 - Check that un-checking the "Include_Points_Coords" option exports CSV without Points values.

### DIFF
--- a/omero/export_scripts/Batch_ROI_Export.py
+++ b/omero/export_scripts/Batch_ROI_Export.py
@@ -120,13 +120,15 @@ def get_export_data(conn, script_params, image, units=None):
                             "mean": stats[0].mean[c] if stats else "",
                             "std_dev": stats[0].stdDev[c] if stats else ""
                         }
+                        if hasattr(image, 'well_id'):
+                            row_data['well_id'] = image.well_id
                         add_shape_coords(shape, row_data,
                                          pixel_size_x, pixel_size_y)
                         export_data.append(row_data)
 
     return export_data
 
-
+# well_id inserted if SPW
 COLUMN_NAMES = ["image_id",
                 "image_name",
                 "roi_id",
@@ -251,7 +253,9 @@ def get_images_from_plate(plate):
     imgs = []
     for well in plate.listChildren():
         for ws in well.listChildren():
-            imgs.append(ws.image())
+            img = ws.image()
+            img.well_id = well.id
+            imgs.append(img)
     return imgs
 
 
@@ -261,6 +265,8 @@ def batch_roi_export(conn, script_params):
 
     dtype = script_params['Data_Type']
     ids = script_params['IDs']
+    if dtype in ("Screen", "Plate"):
+        COLUMN_NAMES.insert(1, "well_id")
     if dtype == "Image":
         images = list(conn.getObjects("Image", ids))
     elif dtype == "Dataset":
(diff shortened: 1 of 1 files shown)


omero/export_scripts/Batch_ROI_Export.py
```diff
@@ -337,8 +364,16 @@ def run_script():
                          "where Z and T are not set?"),
             default=False),
 
+        scripts.Bool(
+            "Include_Points_Coords", grouping="5",
+            description=("Export the Points string for Polygons "
+                         "and Polylines. Disable this to reduce the "
+                         "size of the csv file when exporting large "
```
mtbc:
```suggestion
                         "size of the CSV file when exporting large "
```

mtbc:
Should we also omit the Points heading from `COLUMN_NAMES` if not `include_points`?

mtbc (approved):

will-moore:
Good to merge?

jburel:
I understand that was needed for the prototype.
Was the script tested when there is a large number of ROis and with the "include" points flag on. 
I am wondering if we should not indicate a limit, maybe in the help, before it starts going sideways

will-moore:
The amount of work to decide on a limit (which will depend on your setup as well as the number of columns in the table and the size of the Polygons you're exporting) is not worth it I think.

The failure of this script when exporting large numbers of Polygons/Polylines is not new, and we haven't had anyone reporting problems.

will-moore:
Another option is to export simplified Polygon to preserve the shape but without a long list of coordinates.
See https://shapely.readthedocs.io/en/latest/manual.html#object.simplify
As used at https://docs.google.com/presentation/d/1QzKYP6sXPefBMNfY4PW4H0AMoWVbw9HeNoweLmJg17Y/edit#slide=id.g823da5adf3_0_105

joshmoore:
Might be interesting to also use that as an opportunity to allow exporting WKT (https://shapely.readthedocs.io/en/latest/manual.html#well-known-formats) so we could begin building up familiarity with that.

will-moore:
Trying to get some old PRs merged...
Can re-list if needed? But I think it has been approved already.

jburel:
The PR was approved prior to that discussion and was part of some work on a prototype. So it seems like a good opportunity to test `shapely` as mentioned above

will-moore:
The use of shapely is really unrelated to this PR (adding Well IDs).
It will need a bit of investigation/decision on whether to add the shapely dependency to this script and script parameter options for using it.
I added the comment on this PR since there was already a discussion about Points being too verbose. But that issue is not due to this PR. I have created a separate issue for that at https://github.com/ome/omero-scripts/issues/185

jburel:
The points were previously included by default. To keep the same behaviour the ``include_points`` flag should be set to ``true`` in the various functions so the original behaviour is kept.

dominikl:
Looks good to me. Tested on merge-ci, does what it says 👍

will-moore:
Re-listing for review again - trying to get some more PRs closed...

PR #180, opened 2021-01-21
bom fix populate metadata

### LINKED ISSUE #179: Populate metadata and BOM
The Populate Metadata script fails if the CSV file which is fed into it has a BOM marker.
The BOM marker is easily created by following workflow

1. Open Microsoft Excel. Import (i.e. do not Open) a CSV file. Edit some bits of this file and save it again as CSV (that is how a user would inadvertently obtain a BOM merked file).

2. You can set or unset the using VIM.  (command ``set nobomb`` to unset, ``set bomb`` to set).

Suggestion: Make the Populate Metadata script robust to handle CSV files with BOM. similarly to https://github.com/ome/omero-metadata/pull/40/commits/44f69e9563f2d2c595030a94e7c51b9e21c9f1a6

cc @sbesson @manics 

```
omero_metadata.populate.MetadataError: 
Column type 'ï»¿well' unknown.
Choose from following: plate,well,image,dataset,roi,d,l,s,b
```

will-moore:
Fixes #179 

To test, see ticket above for how to reproduce (create csv with `bom`).

 - Attach csv to Project/Dataset etc and run Populate_Metadata script.
 - Script should work to create bulk_annotations table.

cc @pwalczysko

### DIFF
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -25,12 +25,51 @@ Populate metadata from CSV.
 import omero
 from omero.gateway import BlitzGateway
 from omero.rtypes import rstring, rlong
+from omero.util.temp_files import create_path
 import omero.scripts as scripts
 import omero.model
 
+import tempfile
 import sys
 
-from omero.util.populate_roi import DownloadingOriginalFileProvider
+
+class DownloadingOriginalFileProvider(object):
+
+    """
+    Provides original file data by downloading it from an OMERO raw file store.
+
+    Ported from omero.util.populate_roi to use NamedTemporaryFile instead
+    of TemporaryFile
+    """
+
+    # Default raw file store buffer size
+    BUFFER_SIZE = 1024 * 1024  # 1MB
+
+    def __init__(self, service_factory):
+        self.service_factory = service_factory
+        self.raw_file_store = self.service_factory.createRawFileStore()
+        self.dir = create_path("populate_roi", "dir", folder=True)
+
+    def get_original_file_data(self, original_file):
+        """
+        Downloads an original file to a temporary file and returns an open
+        file handle to that temporary file seeked to zero. The caller is
+        responsible for closing the temporary file.
+        """
+        print("Downloading original file: %d" % original_file.id.val)
+        self.raw_file_store.setFileId(original_file.id.val)
+        temporary_file = tempfile.NamedTemporaryFile(mode='rt+', dir=str(self.dir))
+        size = original_file.size.val
+        for i in range((size // self.BUFFER_SIZE) + 1):
+            index = i * self.BUFFER_SIZE
+            data = self.raw_file_store.read(index, self.BUFFER_SIZE)
+            temporary_file.write(data.decode("utf-8"))
+        temporary_file.seek(0)
+        temporary_file.truncate(size)
+        return temporary_file
+
+    def __delete__(self):
+        self.raw_file_store.close()
 
 try:
     # Hopefully this will import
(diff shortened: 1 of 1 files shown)


joshmoore:
The copy of the class is so that the next version of the scripts doesn't depend on a particular version of omero-py?

will-moore:
Yes. Maybe I'm still thinking of "release of server-side python code is costly", and it's not such a big deal to simply use `NamedTemporaryFile` in the util class, and release `omero-py`.
But is there a way we can make the `omero-scripts` depend on the updated version of `omero-py`?

pwalczysko:
Tested on 

1. Dataset
   - no BOM
   - with BOM
2. Plate
   - no BOM
   - with BOM

Works as expected. Ready to merge fmpov. Thank you @will-moore

joshmoore:
> But is there a way we can make the `omero-scripts` depend on the updated version of `omero-py`?

Not directly, no. To get new scripts one will (almost certainly) need to download a new OMERO.server.zip. In those instructions, we already say to `pip install -U` but we could be more explicit about it. Perhaps the script could catch the import exception and say as much?

will-moore:
I opened https://github.com/ome/omero-py/pull/274 so when that's released, we could revert the duplication here.
Not sure how this script would check for that change. No imports changed etc.
In due course, when we remove the duplicated code from the script, the script could handle the returned file NOT providing a temp.name and return a suitable error ("please upgrade omero-py").

pwalczysko:
> In due course, when we remove the duplicated code from the script, the script could handle the returned file NOT providing a temp.name and return a suitable error ("please upgrade omero-py").

This is kind of a similar topic which was pointed out https://github.com/ome/omero-guide-upload/pull/13#discussion_r561761470 by @manics  - the script is dependent on omero-metadata plugin as well.

pwalczysko (approved):

jburel:
> I opened [ome/omero-py#274](https://github.com/ome/omero-py/pull/274) so when that's released, we could revert the duplication here.
> Not sure how this script would check for that change. No imports changed etc.
> In due course, when we remove the duplicated code from the script, the script could handle the returned file NOT providing a temp.name and return a suitable error ("please upgrade omero-py").

A new version of omero-py has been released

will-moore:
That last commit should detect if `omero-py 5.9.1` is not installed and tell users to upgrade.
Screenshot from running with `omero-py 5.9.0` (upgrade message) then running with `omero-py 5.9.1` (works):

Screenshot 2021-03-25 at 11 27 51

will-moore:
Tests are failing since they appear to still be using `omero-py 5.9.0` or older?

pwalczysko:
> Tests are failing since they appear to still be using `omero-py 5.9.0` or older?

As far as I understand it, the tests are pulling omero-server from https://hub.docker.com/layers/openmicroscopy/omero-server/latest/images/sha256-1947bf3cb3a92171455193c970b6d98cc50027d2f40363b1a94a839c126280f2?context=explore and this image was not updated since 1 month. 
Thus imho it cannot contain the omero-py 5.9.1 from 22 March and is using omero-py 5.9.0. But maybe I am wrong cc @joshmoore

sbesson (approved):
Thanks @will-moore, the change looks good and being able to support both CSV variants is in line with the CLI plugin.

The main downside is the handling of the `omero-py` dependency upgrade as the user will receive an error message while this will need to be fixed at the server installation level. But this issue is unrelated to this PR and will need to be solved more generically. Current best proposal would be to add the infrastructure to be able to manage `omero-scripts` like the other OMERO.server Python dependencies.

PR #182, opened 2021-02-19
Plot Profile Ignore zero length lines

will-moore:
Tiny fix to Plot_Profile script, noticed while testing https://github.com/ome/omero-py/pull/278

To test:
 - "accidentally" create a line of zero length in iviewer by dragging to draw (instead of click-move-click)
 - Run the Plot_Profile script
 - Other lines should be processed as normal


jburel:
to avoid this type of issue, could you add more tests to https://github.com/ome/omero-scripts/blob/develop/test/integration/test_analysis_scripts.py?

jburel:
Thanks for adding the tests

PR #183, opened 2021-02-19
Docs on readthedocs

jburel:
This PR follows up on some work started by @will-moore 
It publishes the doc on readthedocs. 
see https://jburel-omero-scripts.readthedocs.io/en/latest/

The link in the README under ``Usage`` is not working. This is the suggestion for the final link
we could use "ome" too.

Improve layout in readme too see https://github.com/jburel/scripts/tree/docs_sphinx

To test locally c53d456
* Delete a script
* Run ``make html`` in the docs 
* A warning should be printed out in the console


joshmoore:
Screen Shot 2021-02-22 at 08 51 41

certainly looks nice, J-m! Did you notice whether or not there's a warning/error if a script is added but not linked in the docs?

jburel:
There is no warning or error by default. This will require some code in ``config.py`` to scan the scripts directory and compare with the list of entries with ``automodule``

jburel:
if the script is listed using "automodule" and does not exist, there is a warning
```
WARNING: autodoc: failed to import module 'Kymograph'; the following exception was raised:
No module named 'Kymograph'
```
So I need to add a check for the situation when we have a script but no entry in the doc.

docs/conf.py
```diff
@@ -0,0 +1,173 @@
...
+
+# -- Project information -----------------------------------------------------
+
+project = u'omero scripts'
+copyright = u'2021, Open Microscopy Environment'
+author = u'Open Microscopy Environment'
+
+# The full version, including alpha/beta/rc tags
+# The short X.Y version.
+version = '5.6.0'
```
sbesson: This will need to be kept in sync with https://github.com/ome/omero-scripts/blob/develop/setup.py#L76? If so can this be added to https://github.com/ome/omero-scripts/blob/develop/.bumpversion.cfg ?

sbesson (commented):
Looks good in general, one question about the version synchronization.

In terms of the publishing strategy, is the docs built on push? on tags? And is https://omero-scripts.readthedocs.io/en/latest/ the new canonical URL or are we planning to roll our own domain for these docs?

docs/conf.py
jburel: I was thinking about that too. I am going to see if I can unify that work

jburel:
The readthedocs webhook will build when a PR is merged into master that will become the "latest" as it stands
For the url, we can use either the canonical or a domain specific.
It might be easier to the canonical especially if we roll out that strategy across the various repositories

docs/conf.py
jburel: I have prepared things and added info in README
about the release process of that repository
I had to manually fix the version in ``conf.py`` when this is merged. ``bump2version`` will synchronize the version

jburel:
see https://docs.readthedocs.io/en/stable/versions.html

jburel:
Replaced by https://github.com/ome/omero-scripts/pull/187

PR #184, opened 2021-02-24
Populate metadata Image dtype

will-moore:
This allows the `Populate_Metadata` script UI to pick Image types.

To test:
 - Run on https://merge-ci.openmicroscopy.org/web/webclient/?show=image-39715 (csv already attached)

### DIFF
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -137,7 +137,7 @@ def run_script():
     client = scripts.client(
         'Populate_Metadata.py',
         """
-    This script processes a CSV file, using it to,
+    This script processes a CSV file, using it to
     'populate' an OMERO.table, with one row per Image, Well or ROI.
     The table data can then be displayed in the OMERO clients.
     For full details of the supported CSV format, see
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -110,7 +110,7 @@ def populate_metadata(client, conn, script_params):
 
     if data_type == "Image":
         try:
-            from omero_metadata.populate import ImageWrapper
+            from omero_metadata.populate import ImageWrapper    # noqa: F401
         except ImportError:
             return "Please update omero-metadata to support Image type"
     file_ann_id = None
--- a/omero/import_scripts/Populate_Metadata.py
+++ b/omero/import_scripts/Populate_Metadata.py
@@ -137,11 +137,11 @@ def run_script():
     client = scripts.client(
         'Populate_Metadata.py',
         """
-    This script processes a CSV file, attached to a container,
-    converting it to an OMERO.table, with one row per Image or Well.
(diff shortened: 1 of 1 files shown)


sbesson:
I am aware that https://github.com/ome/omero-cli-server/pull/5 contains the ongoing discussion about how to express the dependencies for this Python packages on OMERO.server. However, I wonder whether `setup.py` should not include and `install_requires` block that would capture these requirements as well (in the case of this PR, I assume `omero-metadata>=0.6.0`)

sbesson:
Tested in conjunction with https://github.com/ome/omero-web/pull/264. The script works as expected, taking the input CSV and generating a table alongside the original CSV

Screenshot 2021-02-25 at 16 08 34

Two additional thoughts:

- should the script help be updated to reflect the fact that metadata can be populated for ROIs at the level of the images?
- coming back to the considerations https://github.com/ome/omero-scripts/pull/184#issuecomment-785169937, do we need to handle the case where the scripts would be upgraded by not the underlying library?

will-moore:
We only add `Image` to the types if `from omero_metadata.populate import ParsingContext` passes.
However, if I have installed an older version of `omero-metadata` e.g. `v0.5.1` when Image wasn't supported, I get this error if I try to use an Image:

```
    self.value_resolver = ValueResolver(client, target_object)
  File "/Users/wmoore/Desktop/METADATA/omero-metadata/src/omero_metadata/populate.py", line 338, in __init__
    'Unsupported target object class: %s' % self.target_class)
omero_metadata.populate.MetadataError: Unsupported target object class: <class 'omero.model.ImageI'>
```

I guess we *could* detect the support for `Image` with `from omero_metadata.populate import ImageWrapper`, but that might not be very reliable. It could be renamed or removed.
I don't know if there's a reliable way to get the version number and show "Please upgrade" if it's not recent or latest?

Otherwise, we should at least state on the `omero-metadata` README (and omero-guides if they mention this workflow) that Image/ROIs support was added in `v0.6.0`. Although, a user won't necessarily know what version is installed on their server.

omero/import_scripts/Populate_Metadata.py
```diff
@@ -136,11 +137,11 @@ def run_script():
     client = scripts.client(
         'Populate_Metadata.py',
         """
-    This script processes a CSV file, attached to a container,
-    converting it to an OMERO.table, with one row per Image or Well.
+    This script processes a CSV file, using it to,
```
joshmoore:
```suggestion
    This script processes a CSV file, using it to
```

will-moore:
That last commit tests to see if `ImageWrapper` is supported in the current version of `omero-metadata` if the user has chosen an "Image" dtype.
If not, return a suitable message:

Screenshot 2021-06-16 at 12 32 04

joshmoore (approved):
I like the feedback to the users if it's not installed. (Better would of course be if scripts could have dependencies. One day...)

I could see having the "Please update" message have a red X rather than a green check, but that's not a strong feeling. :+1:

PR #188, opened 2021-06-02
python3 fixes to Images_From_ROIs.py

will-moore:
This fixes python3 issues with generating Big images from ROIs reported at https://forum.image.sc/t/error-in-images-from-rois-py/53411

To test:
 - Draw a Rectangle over 3k x 3k on a big image - Save
 - Run util script Images_From_Rois on the Image
 - Should get a new Big tiled image

### DIFF
--- a/omero/util_scripts/Images_From_ROIs.py
+++ b/omero/util_scripts/Images_From_ROIs.py
@@ -84,9 +84,9 @@ def create_image_from_tiles(conn, source, image_name, description,
         for c in range(0, size_c):
             for z in range(0, size_z):
                 for tile_offset_y in range(
-                        0, ((size_y + tile_height - 1) / tile_height)):
+                        0, ((size_y + tile_height - 1) // tile_height)):
                     for tile_offset_x in range(
-                            0, ((size_x + tile_width - 1) / tile_width)):
+                            0, ((size_x + tile_width - 1) // tile_width)):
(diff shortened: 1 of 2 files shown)


jburel:
I have reenabled the workflow. This was not run.

jburel:
Could the change be captured in a test?

will-moore:
@jburel This is not trivial to add a test for big images, since `testlib create_test_image(size_x=4000, size_y=4000)` fails with `Non-tile based writing unsupported.`.
To create big tiled images via the API, we need something like https://github.com/ome/omero-py/pull/276/files, but there are other issues as described there.
Maybe this is the incentive we need to find a solution... cc @joshmoore

sbesson:
Is there any requirement to use the RawPixelsStore API to create a large pyramidal image? Would a classical pyramidal file import work for the purpose of the test?

will-moore:
populate_metadata tests failing due to OMERO.tables `message = null table as argument`

joshmoore (approved):
In general the Python 3 changes all make sense. Are there leftovers from the "big image" bits that need capturing anywhere?

will-moore:
@joshmoore No, I think https://github.com/ome/omero-py/pull/276/ covers everything outstanding and https://github.com/ome/omero-figure/issues/403 is the open issue for that.

PR #196, opened 2022-02-02
Images from ROIs track xy

will-moore:
From user request to create a Movie from an ROI when the ROI moves in x and y over time.

This PR improves the `Images_from_ROIs.py` script so that the X and Y of the rectangle at each time point are used to crop the image.

To test:
 - Draw a Rectangle in Insight, then right-click and "propagate" it through Time (drag the square across Time axis to cover all the T-indices you wish to).
 
Screenshot 2022-02-03 at 20 09 49

 - Close that dialog, then update the X and Y position of the Rectangle at each time point (so as to track a moving object)
 - Save the ROI (with multiple Rectangles within it). You can do this for multiple ROIs on the Images.
 - Return to the main Insight window, select the Image and click on the Run-script button (green arrow with cogs) and find the `Images_From_ROIs` script (your hierarchy might look a bit different)

Screenshot 2022-02-02 at 08 49 38

 - You can run this script with the default values.
 - A new Image should be created for (each) ROI and the crop at each time-point should respect the position of the Rectangle at that time-point. These will be in a new Dataset (called `from_ROIs` by default).

 - NB: to export each of these Images as a Movie, use the `Make Movie` functionality for each one in turn.

cc @pwalczysko

### DIFF
--- a/omero/util_scripts/Images_From_ROIs.py
+++ b/omero/util_scripts/Images_From_ROIs.py
@@ -453,7 +453,7 @@ def run_script():
         'Images_From_ROIs.py',
         """Crop an Image using Rectangular ROIs, to create new Images.
 ROIs that extend across Z and T will crop according to the Z and T limits
-of each ROI. If NO Z set, use all Z planes. If NO T set, use all T planes.
+of each ROI.
 If you choose to 'make an image stack' from all the ROIs, the script \
 will create a single new Z-stack image with a single plane from each ROI.
 ROIs that are 'Big', typically over 3k x 3k pixels will create 'tiled'
--- a/omero/util_scripts/Images_From_ROIs.py
+++ b/omero/util_scripts/Images_From_ROIs.py
@@ -167,7 +167,8 @@ def get_rectangles(conn, image_id):
             t_end = max(t_indexes) if t_indexes else None
             z_start = min(z_indexes) if z_indexes else None
             z_end = max(z_indexes) if z_indexes else None
-            rois.append((x, y, width, height, z_start, z_end, t_start, t_end, xy_by_time))
+            rois.append((x, y, width, height, z_start, z_end,
+                         t_start, t_end, xy_by_time))
 
     return rois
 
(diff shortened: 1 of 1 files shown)


pwalczysko:
What is the next step if a movie is a desired end-product ? Combine images script ?

pwalczysko:
Bug
Taking the FRAP image in https://workshop.openmicroscopy.org/webclient/?show=image-28669 and **not** checking the Make Image stack checkbox, I still get the image stack.  

Edit: Follow-up on the above: When I do check the Make Image stack checkbox, I do not get image stack, just one single plane.

Marking as a bug, it seems other way around than it should be, or rather, more precisely, in the case where I do not check the checkbox, I would expect many single images, which does not happen either.

pwalczysko:
As discussed with @will-moore , maybe just adjust the description of the script.

the script principally does what we want, it is historically much more elaborate in comparison of with what is needed in a present case.

pwalczysko:
Understood now what the script is doing. It is definitely catering for the new case and the behaviour is improved.

Ready to merge fmpov.

will-moore:
Steps to update this on nightshade server:
 - [ ] merge this PR
 - [ ] tag this repo
 - [ ] open PR to edit training playbook https://github.com/ome/prod-playbooks/blob/master/omero/training-server/playbook.yml to grab the tagged script and upload
 - [ ] merge that PR, test deploy on a training server
 - [ ] open a corresponding PR to add block to https://github.com/ome/prod-playbooks/blob/f124c1ec8e7562cf50a69b04576eb7e407b0ef1f/ome-dundeeomero.yml#L177-L184
 - [ ] merge that PR
 - [ ] decide on time-slot and send e-mail to notify nightshade users via web ui.
 - [ ] run playbook to update

will-moore:
To test the last commit - drag the Rectangle outside the bounds of the image. Script should run without an error.

pwalczysko:
> To test the last commit - drag the Rectangle outside the bounds of the image. Script should run without an error.

Works as described, it does not crash when ROIs are outside of the image in some timepoints. (could also repeat the crash with the previous script)

All seems good to me here.

PR #199, opened 2022-04-24
Add HCS support to Remove_KeyVal.py

will-moore:
See https://forum.image.sc/t/uploading-key-value-pairs-from-csv-files-into-omero-web-for-plates/60202/48

This adds HCS support to the Remove_KeyVal.py, extending the existing logic for Datasaets (where ALL key-value pairs on a Dataset and it's Images are deleted).

 To test:
 - Add Key-Value pairs to a Plate and 1 or more Wells and Images within it
 - Select the Plate and run the `Remove KeyVal` script
 - All Map Annotations on the Plate, Wells and Images should be deleted

cc @abhamacher

### DIFF
--- a/omero/annotation_scripts/Remove_KeyVal.py
+++ b/omero/annotation_scripts/Remove_KeyVal.py
@@ -28,23 +28,18 @@ Created by Christian Evenhuis
 
 from omero.gateway import BlitzGateway
 import omero
-from omero.cmd import Delete2
 from omero.rtypes import rlong, rstring, wrap
 import omero.scripts as scripts
 
-import re
 
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-def RemoveMapAnnotations(conn, obj ):
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-
-    anns = list( obj.listAnnotations())
+def remove_map_annotations(conn, obj):
+    anns = list(obj.listAnnotations())
     mapann_ids = [ann.id for ann in anns
-         if isinstance(ann, omero.gateway.MapAnnotationWrapper) ]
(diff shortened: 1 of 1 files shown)


abhamacher:
Hi Will,

thanks for your fast implementation. I tested the script on a 96-well and a 384-well plate and it worked well on deleting the KV-pairs on wells and images. I tried it also with single images of a plate, which worked too without issues.

Now, that there are scripts for adding and deleting KV-pairs on plates, it would be lovely to have an "export all KV-pairs" too, which is available in the original KV-pair script from Chris Evenhuis for Datasets. But that has probably not the highest priority, the other two scripts were more importing from my point of view.

Thanks!

JensWendt:
Hi, 
tried it on one plate with all 3 levels (plate, well, image) annotated. Plate was not full but only a few wells with 2 images each.
Worked.

sbesson:
The CI failures seem to be largely independent of this PR but it looks like GitHub Actions got (auto)disabled when https://github.com/ome/omero-scripts/pull/193 was merged while the new added script did not meet the codestyle requirement (`flake8`).

Available options are either to fix these issues or exclude the script e.g. via configuration.

will-moore:
@sbesson I fixed all the flake8 issues with `Remove_KeyVal.py`, so maybe if this could be merged, then I could fix flake8 for the other scripts in a separate PR (probably in #195).

joshmoore:
:+1:

sbesson:
Thanks @will-moore and agreed on fixing only the warning associated with the script modified in this PR and handling the other ones separately. @joshmoore beat me to it it seems. Is it worth capturing the remaining flake8 fixes as an issue?

joshmoore:
> Is it worth capturing the remaining flake8 fixes as an issue?

I started to, but with #195 open (and failing) I figured it's essentially a blocker there.

PR #200, opened 2022-05-02
Hcs keyvals from csv, improve delimiter sniffing

abhamacher:
Expanded exception handling of delimiter sniffing. Code from Jens Wendt, slightly modified and tested by myself.


will-moore:
I've added those 2 commits to my PR at https://github.com/ome/omero-scripts/pull/195 (just to keep the discussion in one place).

PR #201, opened 2022-05-10
Add new script Compare_Ome_Xml.py

will-moore:
Compare OME.xml between images of the same name in selected Dataset(s).

Script added under "Export_Scripts".
Message gives info on how many images are compared and any differences.

NB: If you've imported images from NGFF, you'll need to manually rename them to match images imported from original files.

Screenshot 2022-05-10 at 09 56 29

cc @pwalczysko

### DIFF
--- a/omero/export_scripts/Compare_Ome_Xml.py
+++ b/omero/export_scripts/Compare_Ome_Xml.py
@@ -1,5 +1,4 @@
-
-#!/usr/bin/env python
+# !/usr/bin/env python
 # -*- coding: utf-8 -*-
 
 # -----------------------------------------------------------------------------
@@ -17,17 +16,16 @@
 #   You should have received a copy of the GNU General Public License along
 #   with this program; if not, write to the Free Software Foundation, Inc.,
 #   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
-# 
 
 # This script exports the OME.xml from all Images in a Dataset.
 # The OME.xml from Images with the same Name is compared and
 # any differences are shown in the printed log.
 
-from xml.etree import ElementTree as ET
+from xml.etree import ElementTree
 
 import omero.scripts as scripts
 from omero.gateway import BlitzGateway
-from omero.rtypes import rlong, rstring, robject, unwrap
+from omero.rtypes import rlong, rstring
 
 
 def get_image_ome_xml(conn, image_id):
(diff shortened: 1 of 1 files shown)


omero/export_scripts/Compare_Ome_Xml.py
```diff
@@ -0,0 +1,139 @@
...
+    exporter.close()
+    return xml
+
+
+def compare_xml(xml1, xml2):
+    tree1 = ElementTree.ElementTree(ElementTree.fromstring(xml1))
+    tree2 = ElementTree.ElementTree(ElementTree.fromstring(xml2))
+
+    mismatch = False
+    for child1, child2 in zip(tree1.iter(), tree2.iter()):
```
sbesson, line 47: See https://github.com/ome/ZarrReader/pull/31#issuecomment-1123931946 for a potential concern about the diffing of the trees

will-moore: [9213cfe](https://github.com/ome/omero-scripts/pull/201/commits/9213cfecb7f71363b1717513002f7b3b86ae5438) compares Planes only with other planes of the same Z/C/T index.
This doesn't check that there are the same number of Plane elements in both XML docs - It simply stores the Plane in a dict with `Z/C/T` key, and if there's a Plane already stored with that key then we compare them (assumes they are from the 2 different XML docs, but doesn't check that).

For multi-T image (that was previously flagged as non-matching XML) this now passes: https://merge-ci.openmicroscopy.org/web/webclient/?show=image-231218
For multi Z/C/T image, this still fails https://merge-ci.openmicroscopy.org/web/webclient/?show=image-230717
with output like:

```
Comparing planes Plane:1:1:14
-------- ** difference... ** --------------
nodeA {'DeltaT': '3656.5830078125', 'DeltaTUnit': 's', 'ExposureTime': '0.05000000074505806', 'ExposureTimeUnit': 's', 'PositionX': '925.13', 'PositionXUnit': 'reference frame', 'PositionY': '-5910.84', 'PositionYUnit': 'reference frame', 'PositionZ': '-0.499', 'PositionZUnit': 'reference frame', 'TheC': '1', 'TheT': '14', 'TheZ': '1'}
nodeB {'DeltaT': '1766.001953125', 'DeltaTUnit': 's', 'ExposureTime': '0.07999999821186066', 'ExposureTimeUnit': 's', 'PositionX': '925.13', 'PositionXUnit': 'reference frame', 'PositionY': '-5910.84', 'PositionYUnit': 'reference frame', 'PositionZ': '-0.499', 'PositionZUnit': 'reference frame', 'TheC': '1', 'TheT': '14', 'TheZ': '1'}
...----------------------
```

will-moore:
This has served it's purpose - closing until we need it again...

PR #202, opened 2022-05-17
Add Plate support to KeyVal_to_csv.py

will-moore:
I was prompted by the need for test data for https://github.com/ome/omero-scripts/pull/195,
so I have updated KeyVal_to_csv.py to support `Plate` input.
I seem to remember @abhamacher requested this too somewhere?

I wasn't sure whether to export KV pairs from `wells` or `images` for a Plate.
Combining different KV pairs from images and wells could be tricky.
My initial update allows the user to choose, with a boolean input "Export Wells"? for a Plate.

But if we can define how to combine KV pairs from Wells and Images, I could give it a go.

E.g. if you have a Well containing 2 images:
 - Well: `A1`  with KV pairs: `key1: value1, key2: value2`
 -  Image: `field1` with KV pairs: `key1: value10, key2: value20`
 -  Image: `field2` with KV pairs: `key1: value11, key2: value22`

What should the CSV look like? 1 row or 2 rows or even 3?
cc @JensWendt

### DIFF
--- a/omero/annotation_scripts/KeyVal_to_csv.py
+++ b/omero/annotation_scripts/KeyVal_to_csv.py
@@ -27,94 +27,111 @@ import omero
 from omero.gateway import BlitzGateway
 from omero.rtypes import rstring, rlong
 import omero.scripts as scripts
-from omero.model import PlateI, ScreenI, DatasetI
-from omero.rtypes import *
 from omero.cmd import Delete2
 
 import tempfile
 
-import os,sys
 import csv
-import copy
+import os
 
 from collections import OrderedDict
 
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-def GetExistingMapAnnotions( obj ):
-# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+def get_existing_map_annotions(obj):
     ord_dict = OrderedDict()
     for ann in obj.listAnnotations():
-        if( isinstance(ann, omero.gateway.MapAnnotationWrapper) ):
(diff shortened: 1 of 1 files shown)


JensWendt:
Hi @will-moore ,

I think we should orient ourselves at the same standard that the `KV_from_csv` script implements.
For every object with KV-pairs one row.
For every Key one column.
Resulting in this output for your above sample:

well      | image      |  key1     |  key2
--------|-----------|---------|------
A1        |                | value1   | value2
A1        |  field1      | value10 | value20
A1        |  field2      | value11 | value22

In general I assume it is highly unlikely that a well and the "underlying" images have differing KV-pairs.
I personally recommend my users to use the same KV-pairs for both, for a better overview in certain situations.
With the same reasoning it is unlikely that the image names are not unique in a plate and therefore the well name before the image might not be required in a single plate setup. But if we go for multiple plates in one .csv I think the whole "chain" plate-name, well-name and image-name is required.

Nevertheless I think that creating a .csv that would allow the recreation of the same annotation-setup if it immediately gets used for the reverse `KV_from_csv` is the goal.

I hope that helps (although it comes pretty late, sorry)

snoopycrimecop:
Conflicting PR. Removed from build [OMERO-python-superbuild-push#308](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/308/). See the [console output](https://snoopycrimecop-ci.openmicroscopy.org/jenkins/job/OMERO-python-superbuild-push/308/consoleText) for more details.
Possible conflicts:
  - Upstream changes
    - omero/annotation_scripts/KeyVal_to_csv.py

--conflicts

will-moore:
Covered by other scripts now.

PR #203, opened 2022-08-08
Fixed error message output for Batch_Image_Export

muhanadz:
The script gives two choices to the user; to Export Individual Channels OR to Export Meged Image. The user might fail to select either of the options.

Screenshot 2022-08-08 at 4 17 57 pm

This PR fixed the output message to alert the user that they need to chose either of the option.

#### Without PR
Script fails to exit correctly with `TypeError`.
```
  File "./script", line 625, in <module>
    run_script()
  File "./script", line 609, in run_script
    file_annotation, message = batch_image_export(conn, script_params)
TypeError: 'NoneType' object is not iterable
```

#### With PR
The output message displays "Not chosen to save Individual Channels OR Merged Image" and exits correctly.
image

### DIFF
--- a/omero/export_scripts/Batch_Image_Export.py
+++ b/omero/export_scripts/Batch_Image_Export.py
@@ -283,7 +283,7 @@ def batch_image_export(conn, script_params):
 
     if (not split_cs) and (not merged_cs):
         log("Not chosen to save Individual Channels OR Merged Image")
-        return
+        return None, "Not chosen to save Individual Channels OR Merged Image"
 
     # check if we have these params
     channel_names = []


will-moore (approved):
Tested on merge-ci. Looks good.

sbesson:
@will-moore should we get released as `omero-scripts 5.7.1` so that it can be bundled and distributed in the upcoming OMERO.server release?

will-moore:
@sbesson Sounds good. I'll go ahead...

will-moore:
Done: https://github.com/ome/omero-scripts/releases/tag/v5.7.1

PR #214, opened 2024-01-09
Update

jburel:
* Update the version of actions
* Test with updated test infra

### DIFF
--- a/omero/util_scripts/Move_Annotations.py
+++ b/omero/util_scripts/Move_Annotations.py
@@ -87,17 +87,17 @@ def move_well_annotations(conn, well, ann_type, remove_anns, ns):
     existing_well_keys = [get_key(link) for link in existing_well_links]
 
     new_links = []
-    for link in old_links:
-        if get_key(link) in existing_well_keys:
+    for old_link in old_links:
+        if get_key(old_link) in existing_well_keys:
             continue
-        log("    Annotation: %s %s" % (link.child.id.val,
-                                       link.child.__class__.__name__))
+        log("    Annotation: %s %s" % (old_link.child.id.val,
+                                       old_link.child.__class__.__name__))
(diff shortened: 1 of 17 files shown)


will-moore:
@jburel This PR should fix the `font.getsize()` issue reported by @juliomateoslangerak at https://github.com/ome/omero-figure/issues/522#issuecomment-1920999404

jburel:
@will-moore I add to disable the ``test_move_annotations`` test for now. It failed for ``wells``.
I would suggest that we integrate the set of the changes in this PR after review then fix the test in a follow up PR since it is blocking other work

will-moore (approved):
Code changes look good. Tested Split-View Figure on merge-ci and it fixes the `font.getsize()` error and works OK. 👍

omero/util_scripts/Move_Annotations.py
```diff
@@ -76,28 +76,28 @@ def get_key(ann_link, with_owner=False):
...
+    for link in old_links:
+        if get_key(link) in existing_well_keys:
             continue
-        log("    Annotation: %s %s" % (l.child.id.val,
-                                       l.child.__class__.__name__))
+        log("    Annotation: %s %s" % (link.child.id.val,
+                                       link.child.__class__.__name__))
         link = WellAnnotationLinkI()
         link.parent = WellI(well.id, False)
-        link.child = l.child
```
sbesson, line 97: Isn't the test problem related to the fact that you renamed the loop variable as `link` but the logic also defines a different `link` objet?

This call is a no-op for instance

jburel: well spotted. so many changes I missed it. I will check that

jburel: test reactivated and everything is green

PR #216, opened 2024-02-02
Key-value pair scripts enhancement

Tom-TBT:
Hello, 
here are the four scripts around key-value pairs, enhancing three and adding a new one to convert key-value pairs.

`Import_from_csv.py`: Import key-value pairs and tags from a CSV file
`Export_to_csv.py`: Export key-value pairs and tags to a CSV file
`Remove_KeyVal.py`: Delete key-value pairs
`Convert_KeyVal_namespace.py`: Changes the namespace of key-value pairs

Feel free to propose different names for the scripts, parameters, or UI changes.

Current repo for the associated guide: 
https://github.com/German-BioImaging/guide-KVpairs-scripts/


will-moore:
@Tom-TBT I wonder if you could fix the merge conflicts (merge develop branch into yours)?

I just tried the `Convert_KeyVal_namespace.py` and it works well.
One question I have is whether it might be preferable to update the namespace on existing annotations instead of deleting them and creating new ones? By updating existing ones, you preserve the history (don't change the 'Linked on' date) and it just feels more appropriate for this change.

Just tried modifying the script to do this, which allows you to simplify the code a bit too (see diff):

<details><summary>diff</summary>

```
+++ b/omero/annotation_scripts/Convert_KeyVal_namespace.py
@@ -131,11 +131,11 @@ def main_loop(conn, script_params):
         for target_obj in target_iterator(conn, source_object,
                                           target_type, is_tag):
             ntarget_processed += 1
-            keyval_l, ann_l = get_existing_map_annotions(target_obj,
+            ann_l = get_existing_map_annotions(target_obj,
                                                          old_namespace)
-            if len(keyval_l) > 0:
-                annotate_object(conn, target_obj, keyval_l, new_namespace)
-                remove_map_annotations(conn, target_obj, ann_l)
+            if len(ann_l) > 0:
+                # annotate_object(conn, target_obj, keyval_l, new_namespace)
+                update_map_annotations(conn, ann_l, new_namespace)
                 ntarget_updated += 1
                 if result_obj is None:
                     result_obj = target_obj
@@ -149,48 +149,38 @@ def main_loop(conn, script_params):
 
 
 def get_existing_map_annotions(obj, namespace_l):
-    keyval_l, ann_l = [], []
+    ann_l = []
     forbidden_deletion = []
     for namespace in namespace_l:
         p = {} if namespace == "*" else {"ns": namespace}
         for ann in obj.listAnnotations(**p):
             if isinstance(ann, omero.gateway.MapAnnotationWrapper):
                 if ann.canEdit():  # If not, skipping it
-                    keyval_l.extend([(k, v) for (k, v) in ann.getValue()])
                     ann_l.append(ann)
                 else:
                     forbidden_deletion.append(ann.id)
     if len(forbidden_deletion) > 0:
         print("\tMap Annotation IDs skipped (not permitted):",
               f"{forbidden_deletion}")
-    return keyval_l, ann_l
+    return ann_l
 
 
-def remove_map_annotations(conn, obj, ann_l):
+def update_map_annotations(conn, ann_l, new_namespace):
     mapann_ids = [ann.id for ann in ann_l]
 
     if len(mapann_ids) == 0:
         return 0
-    print(f"\tMap Annotation IDs to delete: {mapann_ids}\n")
+    print(f"\tMap Annotation IDs to update: {mapann_ids}\n")
     try:
-        conn.deleteObjects("Annotation", mapann_ids)
+        for map_ann in ann_l:
+            map_ann.setNs(new_namespace)
+            map_ann.save()
         return 1
     except Exception:
         print("Failed to delete links")
         return 0
 
 
-def annotate_object(conn, obj, kv_list, namespace):
-
-    map_ann = omero.gateway.MapAnnotationWrapper(conn)
-    map_ann.setNs(namespace)
-    map_ann.setValue(kv_list)
-    map_ann.save()
-
-    print("\tMap Annotation created", map_ann.id)
-    obj.linkAnnotation(map_ann)

```
</summary>

You *could* apply this approach to the other scripts too, but that's a bigger change, particularly since they sometimes combine KVP annotations etc too?

Tom-TBT:
OK for the change to "Convert KV namespace". 
Its one advantage was to merge key-value pairs with the same namespace into a new key-value pair. 
What do you think, should it be a fifth script or an option of the convert script? I remember someone, maybe Petr, mentioning that merging key-value pairs was also a desired functionality.

I'll merge develop into my branch too

omero/annotation_scripts/Convert_KeyVal_namespace.py
```diff
@@ -0,0 +1,333 @@
...
+        print(f"\t- {target_obj}")
+        yield target_obj
+
+
+def main_loop(conn, script_params):
+    """
+    For every object:
+     - Find annotations in the namespace
+     - Remove annotations with old namespace
+     - Create annotations with new namespace
```
will-moore: This logic is the behaviour if `merge` is True, but otherwise we just update the ns, right?

omero/annotation_scripts/Export_to_csv.py
```diff
@@ -0,0 +1,582 @@
...
+            P_INCL_NS, optional=False,
+            grouping="2.3",
+            description="Check to include the annotation namespaces" +
+                        " in the csv file.", default=False),
+
+        scripts.Bool(
+            P_INCL_TAG, optional=False,
+            grouping="2.4",
+            description="Check to include tags in the csv file.",
+            default=False),
```
will-moore, line 639: For all these `Bool` parameters, the `optional=False` adds a red asterisk to the UI, but they are still optional (since I have to ignore them if I want to enter a False value). So the `optional=False` is kinda misleading/confusing.

omero/annotation_scripts/Export_to_csv.py
```diff
@@ -0,0 +1,582 @@
...
+
+    separators = [";", ",", "TAB"]
+    # Here we define the script name and description.
+    # Good practice to put url here to give users more guidance on how to run
+    # your script.
+    client = scripts.client(
+        'Export to CSV',
+        """
+    This script exports for the selected objects their name, IDs and associated
+    key-value pairs.
```
will-moore: Could add "Tags" to this description now?

Tom-TBT: added to export and import scripts description

omero/annotation_scripts/Export_to_csv.py
Tom-TBT: set optional to false for all bool parameters

omero/annotation_scripts/Convert_KeyVal_namespace.py
Tom-TBT: Yes that's the behaviour with merge=True, otherwise it changes the namespace (default is just change namespace). 
I changed the description to reflect that

omero/annotation_scripts/Import_from_csv.py
```diff
@@ -0,0 +1,861 @@
...
+        namespaces = [el.strip() for el in rows[0]]
+        namespaces = [ns if ns else NSCLIENTMAPANNOTATION for ns in namespaces]
+        rows = rows[1:]
+    header = [el.strip() for el in rows[0]]
+    rows = rows[1:]
+
+    if not import_tags:
+        idx_l = [i for i in range(len(header)) if header[i].lower() != "tag"]
+        header = [header[i] for i in idx_l]
+        namespaces = [namespaces[i] for i in idx_l]
```
will-moore: I got an error here:
```
  File "/opt/omero/server/omero/tmp/omero_omero-server/510/processo47by84e.dir/./script", line 390, in <listcomp>
    namespaces = [namespaces[i] for i in idx_l]
IndexError: list index out of range
```
I chose not to import tags, and I guess that `namespaces` is an empty list at this point since there's no `namespace` in the csv.

omero/annotation_scripts/Import_from_csv.py
```diff
@@ -0,0 +1,861 @@
...
+    if "<PARENTS>" in to_exclude:
+        to_exclude.remove("<PARENTS>")
+        to_exclude.extend(["PROJECT", "DATASET", "SCREEN",
+                           "PLATE", "RUN", "WELL"])
+
+    params[P_EXCL_COL] = to_exclude
+
+    assert (params[P_CSVSEP] is None
+            or params[P_CSVSEP] not in params[P_SPLIT_CELL]), (
+                "Cannot split cells with a character used as CSV separator"
```
will-moore: I actually have a csv file where commas are used to separate multiple values in each cell. See https://zenodo.org/records/7308444/files/Davis%20lab%20200%20YFP%20trap%20expression.csv
This is possible because each cell in the csv is escaped with double quotes: `one,"twoA,twoB",three`.
So this assertion is a bit of a problem in this case.

omero/annotation_scripts/Import_from_csv.py
Tom-TBT: thank you

omero/annotation_scripts/Import_from_csv.py
Tom-TBT: I removed that problematic assertion. Thanks again

omero/annotation_scripts/Import_from_csv.py
```diff
@@ -0,0 +1,858 @@
...
+            default="OBJECT_NAME",
+            description="The column name in the .csv containing the name of " +
+                        "the objects to annotate (used if no column " +
+                        "ID is provided or  found in the .csv). Matches " +
+                        "<NAME> in exclude parameter."),
+
+        scripts.Bool(
+            P_ATTACH, optional=True, grouping="3.7", default=False,
+            description="Attach the given CSV to the selected objects" +
+            "when not already attached to it."),
```
will-moore: missing whitespace `objectswhen`

omero/annotation_scripts/Import_from_csv.py
```diff
@@ -0,0 +1,858 @@
...
+                parsed_ns, exclude_empty_value, tagid_d, split_on
+            )
+
+            if updated:
+                if result_obj is None:
+                    result_obj = target_obj
+                ntarget_updated += 1
+                ntarget_updated_curr += 1
+
+        if ntarget_updated_curr > 0 and attach_file:
```
will-moore: I'm not sure of the logic here, but I chose to attach the file, then ran the script but forgot to choose the correct target so it "failed" didn't get attached.
Each time this happens we end up with an orphaned FileAnnotations since it doesn't get deleted. And the user can't delete Orphaned FileAnnotations without attaching them first (at least in the webclient).
I would almost consider *always* attaching the FileAnnotations, unless there is a particular reason to give the user a choice?

omero/annotation_scripts/Import_from_csv.py
Tom-TBT: Fixed this one and others from all four scripts

omero/annotation_scripts/Import_from_csv.py
Tom-TBT: That was a misunderstanding from me, I thought those were considered temporary files. The logic was to not attach the file if it doesn't work, to avoid the extra step of deleting the "bad csv". I don't see an issue for removing that choice. Now with all other parameters, it will clear the script a little. 

But the FileAnnotation could be deleted too, right? 
Thinking about not having the permissions to annotate an object (we should clean up the uploaded csv)

Tom-TBT:
Hi @will-moore,
I'm taking care of the file attachement issue today. I did the test and like you said, the FileAnnotation remains on the server, no matter if it's attached or not.

Looking at it closely, I also noticed that I should use a better namespace.
I saw that other scripts use such NS for their FileAnnotation: `openmicroscopy.org/omero/scripts/results/created/omero/export_scripts/Batch_Image_Export`

Is it ok if I use the following two from now on:
* `openmicroscopy.org/omero/scripts/results/created/omero/annotation_scripts/Import_from_CSV`
* `openmicroscopy.org/omero/scripts/results/created/omero/annotation_scripts/Export_to_CSV`

Tom-TBT:
@will-moore  Could you please trigger the workflow? I have tests for all four scripts, and flake8 should be fine this time.

joshmoore:
(done)

test/integration/test_annotation_scripts.py
```diff
@@ -0,0 +1,657 @@
...
+        assert value[2] == ("key_3", "val_F")
+
+        value = list(well_a3.listAnnotations(ns=ns))[0].getValue()
+        assert len(value) == 3
+        assert value[0] == ("key_1", "val_G")
+        assert value[1] == ("key_2", "val_H")
+        assert value[2] == ("key_3", "val_I")
+
+    @pytest.mark.parametrize('import_tag', [True, False])
+    @pytest.mark.parametrize('tag_creation', [False])
```
will-moore: Did you want to test with `tag_creation` as `True` too?

Also wondered in `test_import()` above whether you wanted to check for a different outcome with `tag_creation` is True, or if this is just parameterized to check that the script still passes with True and False?

test/integration/test_annotation_scripts.py
```diff
@@ -0,0 +1,657 @@
...
+        plates = self.import_plates(client, plate_cols=n_well, plate_rows=1)
+        plate = plates[0]
+
+        cvs_file = create_path("test_kvp_name", ".csv")
+        # create a file annotation
+        with open(cvs_file.abspath(), 'w') as f:
+            f.write("OBJECT_NAME; key_1; key_2\n")
+            f.write("A1; val_A,val_B; val_C\n")
+            f.write("A2; val_D,val_E,val_F;\n")
+            f.write("A3; ; val_G,val_H\n")
```
will-moore, line 268: Surprising to me that with this CSV the separator is guessed to be `;` but I suppose you have slightly more `;` than you have commas so 👍

will-moore: I had to read the test code quite carefully to understand that `test_import_split` was testing the splitting of a single CSV value into two KVPs, so maybe a doc string on the test would help.

test/integration/test_annotation_scripts.py
```diff
@@ -0,0 +1,657 @@
...
+        kv = client.sf.getUpdateService().saveAndReturnObject(kv)
+        self.link(image, kv, client=client)
+
+        args = {
+            "Data_Type": rstring("Image"),
+            "IDs": rlist([rlong(image.id.val)]),
+            "Target Data_Type": rstring("<on current>"),
+            "Old Namespace (blank for default)": rlist([rstring("test")]),
+            "New Namespace (blank for default)": rstring("new_ns"),
+            "Create new and merge": rbool(True)
```
will-moore: This test looks almost identical to the one above apart from this bool - I wonder if it makes sense to combine the tests and parametrize on that?

test/integration/test_annotation_scripts.py
```diff
@@ -0,0 +1,657 @@
...
+        kv.setNs(rstring("test_delete"))
+        kv = client.sf.getUpdateService().saveAndReturnObject(kv)
+        self.link(image, kv, client=client)
+
+        args = {
+            "Data_Type": rstring("Image"),
+            "IDs": rlist([rlong(image.id.val)]),
+            "Target Data_Type": rstring("<on current>"),
+            "Namespace (blank for default)": rlist([rstring("test_delete")]),
+            agreement: rbool(True)
```
will-moore: parametrize on this param to be sure that deletion doesn't happen without agreement?

will-moore (approved):
Some minor points/comments etc from reading the tests but nothing here is a blocker. 

To test everything is pretty tricky and these seem to give reasonable coverage of the main usage.

test/integration/test_annotation_scripts.py
Tom-TBT: Only the last parts are different, so it is not an issue to merge them. Thanks

test/integration/test_annotation_scripts.py
Tom-TBT: Good point

test/integration/test_annotation_scripts.py
Tom-TBT: Yes, that's a miss from when I was debugging my script. Thanks.
The test above was to check if it passes in both cases (there are no columns in the CSV with tags)

test/integration/test_annotation_scripts.py
Tom-TBT: I added docstring to all tests.

Tom-TBT:
@will-moore  please approve the workflow. I adjusted the script to your suggestions. Thanks for the review

Tom-TBT:
Hello @will-moore @jburel,
is there anything left blocking the merge of this PR? 

Cheers

ping @pwalczysko

pwalczysko:
I was thinking we could use this on our next outreach cc @jburel

omero/annotation_scripts/README.md
```diff
@@ -7,43 +7,38 @@ This is the central repository for community contributed scripts to [omero-web](
...
-icon in the  menu bar.
+You should also configure the Export_to_csv script so that it returns the csv file as a direct download link:
+https://guide-kvpairs-scripts.readthedocs.io/en/latest/setup.html#configuring-the-export-script
 
 History
 --------
 This repository started as a fork of [evehuis/omero-user-scripts](). Ownership was transferred to @CFGrote after merging a pull request that fixed a number of bugs and
 ported the original code from python2.7 to python3.x
 
+In 2023, the scripts were reworked by Tom Boissonnet and Jens Wendt to extend the annotation to all OMERO objects, and to include a new script to convert namespaces of map annotations.
```
jburel: Maybe 2024

Tom-TBT: I changed the date

jburel:
@Tom-TBT you should add yours and @JensWendt institution e.g. Gerbi when indicating your name

will-moore (approved):
All good from me, thanks!

Tom-TBT:
I added our institute affiliations

Tom-TBT:
Anything else from you @jburel? Sorry for being pushy, I'm trying to close that topic.

joshmoore:
One heads up in terms of versioning: the removal or renaming of scripts in my mind would count as a breaking change.

will-moore:
So it would mean a `v5.9.0` tag once this is merged?

sbesson:
> One heads up in terms of versioning: the removal or renaming of scripts in my mind would count as a breaking change.

Seconding that. In my mind that would almost justify calling the next release of `omero-scripts` as `v6.0.0`.

Semi-related, is there any reference documentation about the existing scripts that would need to be updated? And for end-users who would have been using the previous key/value scripts in production, are there  simple upgrade recommendations?

Tom-TBT:
About documentation, I have made documentation for the scripts, that lives on its own for now. The plan was to merge it with your doc after the merge of this branch is completed, if I recall correctly what JM said.

The repo:
https://github.com/German-BioImaging/guide-KVpairs-scripts

The doc:
https://guide-kvpairs-scripts.readthedocs.io/en/latest/index.html

Rdornier:
Hello, 

Apologize for entering the game so late. I've tested the scripts and I found them very good !

I have a suggestion regarding the csv file that is generated. Would it be possible to add the separator used when it is generated (by adding in the first line `sep=%separator%` ) ?  

This will ease the csv file opening and saving with Excel. For example, I have an Excel configured to recognize commas as separator. If the image names don't contain any comma, that's fine but if they contain some, it becomes a mess. I already give those scripts to some users and I made them aware of such issues but it is always better to force reading the file with the correct separator.

I hope it's not too late for that.

Rémy.

Tom-TBT:
Hi @Rdornier,
I was not aware that such magic was possible for Excel. I simply thought it was poorly written because with Open Office (on my home computer) I can open CSV files without this problem.

The annoying trick I used with Excel is to open a blank sheet and in `Data -> From Text/CSV` select your CSV file. Only then Excel is capable of detecting the separator. So I agree to have it easier.

Two issues I can fix:
* writing the CSV better (quotes around the text so commas are fine with the export of comma as separator)
* using that metadata in the first line. Should be minor changes

Thanks a lot for the feedback. That shouldn't be too late

Rdornier:
> I was not aware that such magic was possible for Excel

I didn't know neither until today... I found this [on stackoverflow](https://stackoverflow.com/questions/46555818/change-default-delimiters-in-excel)

Rdornier:
Hi @Tom-TBT

I have another feedback for you. When I try to convert a namespace from `default_namspace` to `A_namespace`, with the checkbox to create new and merge activated, it converts the KVP namespace correctly but doesn't merge the two KVPs together (before running the script, some KVPs were already existing under  `A_namespace`).

To be able to merge the two KVPs in one, I had to run again the script, convert from  `A_namespace` to  `A_namespace` and check the box.
I think it is better if we can remove the extra run.

Thanks,
Rémy.

Tom-TBT:
Oh, I see. The extra run could currently be avoided by giving to "Old Namespace" the two namespaces as input:
`openmicroscopy.org/omero/client/mapAnnotation, A_namespace`

But ticking "create new and merge" seems like an explicit agreement to merge all key-value pairs into a single A_namespace. When ticking that box, I can automatically add the "New namespace" to the "Old namespace" list.

Rdornier:
> The extra run could currently be avoided by giving to "Old Namespace" the two namespaces as input:
> openmicroscopy.org/omero/client/mapAnnotation, A_namespace

Ok , it works !

> But ticking "create new and merge" seems like an explicit agreement to merge all key-value pairs into a single A_namespace. When ticking that box, I can automatically add the "New namespace" to the "Old namespace" list.

Yes, checking the box to bypass the need of writing both namespaces would be very good to have.
Thanks !

Tom-TBT:
@Rdornier   I included both of your suggestions. You can go ahead and try them out with your users. Let me know if that solves your issues.

Rdornier:
@Tom-TBT Thank you very much for your changes and for your reactivity ; I've just tested them and it works perfectly on my side ! I'm pretty sure it will also work for the users.

pwalczysko:
@Tom-TBT see https://github.com/ome/omero-scripts/pull/216#issuecomment-2429933200 please - not sure if you want to change the script at this stage, but the use case mentioned is clearly asking for more robust options. Such as, imho, to not to have select the KVPs on Images or Datasets only in the options of the ``Export_to_csv.py`` script, but to have also an option which would export both types of KVPs into one CSV in one run of the script. In any case, it just shows how useful this set of script really is, so thank you.

Rdornier:
Hi @Tom-TBT ,

I recently had a small issue with the export script when trying to export from an entire plate.
This was coming from the fact that there were no PlateAcquisition object for the plate (see the first screen)

image

The error I got was `NoneType has no attribute OMERO_CLASS` . I fixed it by adding a None check on the `PlateAcquitision` object (l226-227)
```
 if o is not None:
      ancestry.append((o.OMERO_CLASS, get_obj_name(o)))
```

Rémy

Tom-TBT:
Thanks @Rdornier, I could reproduce and your report helps a lot.

The whole code snippet is 
```python
for o in target_obj.getAncestry():
    if o.OMERO_CLASS == "WellSample":
        o = o.getPlateAcquisition()
    ancestry.append((o.OMERO_CLASS, get_obj_name(o)))
```
The "None" Run is not a result of the ancestry iteration but comes from my if condition. So I'll limit the patch to that condition.

Rdornier:
Hi @Tom-TBT 

I find out small improvements that can be done in the `Import_fron_csv` script 
- There is a missing separator in the [list of available separator](https://github.com/ome/omero-scripts/pull/216/files/a48786fff1a4ff6be52015d3a5604f8299ecccaf#diff-a9895d8ba548baefeb78e0f5e4abbb33c60a01eae15f5d39922532599e8388bfR660) ; `|` is missing (the export script gives it as a separator).
- The script was failing when the csv file was resaved with LibreOffice on Linux, due to the `"sep="`  that is not matching [the regex](https://github.com/ome/omero-scripts/pull/216/files/a48786fff1a4ff6be52015d3a5604f8299ecccaf#diff-a9895d8ba548baefeb78e0f5e4abbb33c60a01eae15f5d39922532599e8388bfR368) for `sep=`. Quotes were apparently added around this token. To make it work, I think you could modify the regex like this `\"?sep=(?P<delimiter>.?)\"?`

Rémy.

Tom-TBT:
@Rdornier  Thanks as always for the tests and debugging. I added the missing separator in the option list for import, and the regex can now handle all `"sep=,"` and `'sep=,'` and `sep=,`

pwalczysko:
For OMERO.guides, please consider opening a PR https://github.com/ome/omero-guide-scripts - this is to have a new menu item in https://omero-guides.readthedocs.io/en/latest/scripts/docs/index.html @Tom-TBT - sorry, during the call I erroneously pointed you to ``...introduction``

will-moore:
Just re-capping the changes that users will see to webclient scripts UI...

BEFORE this PR we get this under `annotation_scripts`:

Screenshot 2024-12-02 at 09 52 40

and AFTER we get:

Screenshot 2024-12-02 at 13 13 12

`Key_Val` is removed from the script names since they also handle other annotations (Tags) now.


Comparing similar scripts BEFORE:

Screenshot 2024-12-02 at 09 49 43

and AFTER: we get a lot more parameters but also a link to the guide at https://guide-kvpairs-scripts.readthedocs.io/en/latest/index.html

Screenshot 2024-12-02 at 13 14 51

Similar changes to the export script BEFORE:

Screenshot 2024-12-02 at 13 26 26

and AFTER (showing increased support for more Object types):

Screenshot 2024-12-02 at 13 13 51

So, to conclude, users will be directed to docs/guides for the new scripts, once they pick the new script names from the menu. More complex options for the scripts may take some time to learn, but the "default" usage shouldn't be too hard to find.
NB: If the guides are moved from that location https://github.com/German-BioImaging/guide-KVpairs-scripts/ into OME's OMERO guides, those links will need updating accordingly.

pwalczysko:
In line with the effort on the ``omero-guide-scripts`` we are targeting merge&release on Monday

will-moore:
Once https://github.com/ome/omero-guide-scripts/pull/3 is merged, would we expect the guide to be at
https://omero-guides.readthedocs.io/en/latest/scripts/docs/annotation_scripts.html ?

If so, then in this PR, all the https://guide-kvpairs-scripts.readthedocs.io/en/latest/index.html links should be replaced with that URL?

Tom-TBT:
@will-moore @pwalczysko  I changed the links. If we see that the links are wrong tomorrow, we can open a PR to fix them.

Tom-TBT:
@pwalczysko @will-moore   The links are correct. This can be merged

will-moore (approved):
Links to guides are looking good. LGTM

cc @jburel

PR #217, opened 2024-02-23
Future removal

jburel:
* Check build on 3.8 to 3.12
* Remove future and try/Catch block
* Remove Pytest command

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -4,19 +4,27 @@ on: push
 
 jobs:
   build-n-publish:
+    strategy:
+      matrix:
+        python-version:
+          - '3.8'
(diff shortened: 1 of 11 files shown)


sbesson (commented):
All cleanup changes (future, PIL/Pillow) look consistent with what has been done at the omero-py & omero-web level. Looking forward to seeing whether this introduces regressions in the CI builds.

On the changes to the GitHub action, this workflow only builds the source distribution and a wheel of the omero-scripts. Although testing different versions cannot hurt, it is less instructive than other Python components where the workflow also includes some testing.

jburel:
I agree that it only tests the build. To test the full stack, changes will be required to test-infra

sbesson (approved):

PR #221, opened 2024-06-20
Fix regression introduced in 6923640

ehrenfeu:
Python3 doesn't have `StringTypes` any more, currently the import fails due to the changes introduced in commit 6923640.


will-moore (approved):
Looks good, thanks

sbesson (approved):

PR #230, opened 2025-11-13
Import from csv fix

### LINKED ISSUE #224: Import_from_csv script : PlateAcquisition parent added as KVPs
Hello @Tom-TBT , 

While using the `Import_from_csv` script on a screen, I noticed that the `Plateaquisition` parent header is automatically added in the list of Key-value pairs to add which, I think, should not be the case.

Image

I think this is because you exclude the column `Run` instead of `PlateAcquisition` in that [line](https://github.com/ome/omero-scripts/blob/develop/omero/annotation_scripts/Import_from_csv.py#L983).

Rémy.

Rdornier:
@Tom-TBT @will-moore 

Fix #224

### DIFF
--- a/omero/annotation_scripts/Import_from_csv.py
+++ b/omero/annotation_scripts/Import_from_csv.py
@@ -980,7 +980,7 @@ def parameters_parsing(client):
     if "<PARENTS>" in to_exclude:
         to_exclude.remove("<PARENTS>")
         to_exclude.extend(["PROJECT", "DATASET", "SCREEN",
-                           "PLATE", "PLATEACQUISITION", "WELL"])
+                           "PLATE", "RUN", "PLATEACQUISITION", "WELL"])
 
     params[P_EXCL_COL] = to_exclude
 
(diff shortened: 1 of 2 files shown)


Tom-TBT:
Ola,
I think you can leave RUN in the exclude key, that's an alias used around for PLATEACQUISITION. 
There's always the case someone uses RUN in their csv.

And I think you mixed up with changes on the Split_View script for which you made another PR.

Cheers,
Tom

Rdornier:
Hello Tom,

Thanks for reviewing !

> And I think you mixed up with changes on the Split_View script for which you made another PR.

Arrh, you're right. I've fixed it, thanks

Tom-TBT (approved):
This works as expected, lgtm

sbesson:
I assume https://omero-guides.readthedocs.io/en/latest/scripts/docs/parameter_description.html#include-parent-container-names should be updated accordingly if we get this included in the next OMERO.server release ? /cc @jburel @pwalczysko

Rdornier:
Hi @sbesson, Yes, you're right. This is now done. Thanks!

PR #231, opened 2025-12-04
Create Release 

jburel:
Create a release automatically when a new tag is pushed

### DIFF
--- a/.github/workflows/publish_pypi.yml
+++ b/.github/workflows/publish_pypi.yml
@@ -7,8 +7,6 @@ jobs:
     strategy:
       matrix:
         python-version:
-          - '3.8'
-          - '3.9'
           - '3.10'
           - '3.11'
           - '3.12'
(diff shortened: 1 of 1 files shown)


sbesson (approved):
This is consistent with the strategy I have been applying across several repositories including OMERO.py but also Java components like `ome-common`

Auto-generating the release notes for individual components is fairly low-maintenance and generally produces a good outcome. I also find this approach to be complementary of writing more curated release notes e.g. for Bio-Formats and OMERO as the latter tend to be more user-facing.

Two things to keep in mind while pivoting towards this strategy:

- there is no single entry point that aggregates the changelog of all changes across versions. The closest equivalent would be the  releases endpoint e.g.https://github.com/ome/omero-scripts/releases but this might involve some scrolling & navigation if looking at the change across several releases
- the auto-generation of the release body further increases the importance of good titles (& bodies) for the individual contributions e.g. avoiding generic titles like `Fix bug`. This puts additional responsibility on maintainers but also confers them with further authority during the review process to ensure these are correctly filled.

